PR #1, opened 2020-07-14
Default to prepare_v2 if prepare_v3 not found

alextekartik:
So it seems that travis is still stuck on Ubuntu 16.04 causing the library to fail (hence all the sqflite tests - I added an issue here https://travis-ci.community/t/change-ubuntu-distribution-from-16-04/9320 but I'm not sure when/if it will get fixed).

Here I propose to try to load prepare_v3 and load prepare_v2 if not found.

I also added travis tests (see the build history https://travis-ci.org/github/tekartikdev/sqlite3.dart/builds to see that it was failing before - I made 2 attemps to fix the issue). You might want to change the implementation.

Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: dart
+sudo: required
+dart:
+  - stable
+  - beta
+  - dev
+before_script:
+  - sudo apt-get -y install libsqlite3-dev
+script:
+  - cd sqlite3
+  - pub get
+  - pub run test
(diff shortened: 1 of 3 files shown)


simolus3:
Thanks for the PR! Unfortunately the `.ffi.dart` file is auto-generated, so your changes would be overridden. With the linked commit, we'll just look those functions up outside of that file. I also made the CI run on Ubuntu 16.04.

alextekartik:
Ah ok sorry for missing that. I guess I just throw something to see that this was indeed the only remaining issue (as I could not test it locally). Great indeed that you test on ubuntu 16.04. Thanks! Looking forward for a new published version!

simolus3:
Just published version `0.1.4` which should contain this logic.

alextekartik:
Hooray! : https://travis-ci.org/github/tekartik/sqflite/builds/708061710

I published sqflite_common_ffi 1.1.0 based on your lib. Thanks again for your work. Bravo!

PR #15, opened 2020-12-02
Fix dynamic library usage on iOS

kuhnroyal:
* attempt to run integration test on an older iOS version as well

### DIFF
--- a/sqlite3/lib/src/load_library.dart
+++ b/sqlite3/lib/src/load_library.dart
@@ -44,7 +44,7 @@ DynamicLibrary _defaultOpen() {
     }
   }
   if (Platform.isIOS) {
-    return DynamicLibrary.process();
+    return DynamicLibrary.open('sqlite3.framework/sqlite3');
   }
   if (Platform.isMacOS) {
     return DynamicLibrary.open('/usr/lib/libsqlite3.dylib');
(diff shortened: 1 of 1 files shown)


kuhnroyal:
There doesn't seem to be an easy way of running the integration tests on older simulators with Github actions.

simolus3:
Thank you for this fix :+1: I don't think we need to run tests on older simulators, we can check the compile-time options in the test output.
I think it also makes sense to fallback to `DynamicLibrary.process()` if we can't load the proper library (you can catch the `ArgumentError`). That would also help for people using the sqlcipher variant.

kuhnroyal:
Ok, I can do that.
I don't see a difference in the compile-time options compared to the latest master build.

kuhnroyal:
After looking at sqlcipher_flutter_libs I am actually not sure why `process()` works there (it is 3.33.0). I assume that it might be statically linked due to the definition of `"HEADER_SEARCH_PATHS": "$(PODS_ROOT)/SQLCipher"` in its podspec.

kuhnroyal:
`sqlite3_flutter_libs` should in theory also work for MacOS, not sure if the podspec needs changes for that but the same fallback mechanism could be applied there as well.

simolus3:
For macOS support, I think we'd need another entry in the pubspec and a `macos` folder similar to the current `ios` folder in `sqlite_flutter_libs` too.

PR #30, opened 2021-04-28
applyWorkaround and openCipherOnAndroid

### LINKED ISSUE #29: android 6.1 for sqlcipher_flutter_libs
i see there is error with dynamic link in old phone and some 6.1 device
there is `applyWorkaroundToOpenSqlite3OnOldAndroidVersions` for sqlite3_flutter_libs but not for sqlcipher_flutter_libs
also when we override like this as describe in moor docs  will break old device  because we need to use full path with id on old device 
```dart
  open.overrideFor(
      OperatingSystem.android, () => DynamicLibrary.open('libsqlcipher.so'));
```

i suggest to add this to sqlcipher_flutter_libs : 

1- applyWorkaroundToOpenSqlite3OnOldAndroidVersions 

2- provide openCipherOnAndroid function that do this 
```dart
DynamicLibrary openCipherOnAndroid() {
  try {
    return DynamicLibrary.open('libsqlcipher.so');
  } catch (_) {
    // On some (especially old) Android devices, we somehow can't dlopen
    // libraries shipped with the apk. We need to find the full path of the
    // library (/data/data/<id>/lib/libsqlcipher.so) and open that one.
    // For details, see https://github.com/simolus3/moor/issues/420
    final appIdAsBytes = File('/proc/self/cmdline').readAsBytesSync();

    // app id ends with the first \0 character in here.
    final endOfAppId = max(appIdAsBytes.indexOf(0), 0);
    final appId = String.fromCharCodes(appIdAsBytes.sublist(0, endOfAppId));

    return DynamicLibrary.open('/data/data/$appId/lib/libsqlcipher.so');
  }
}
```
then programmer could do 
```dart
  open.overrideFor(
      OperatingSystem.android, openCipherOnAndroid);
```

i am happy to make pull request if this way ok .

Ali1Ammar:
fixed #29 

add applyWorkaroundToOpenSqlite3OnOldAndroidVersions

add openCipherOnAndroid function 

update readme

### DIFF
--- a/sqlcipher_flutter_libs/README.md
+++ b/sqlcipher_flutter_libs/README.md
@@ -23,6 +23,6 @@ There appears to be a problem when loading native libraries on Android 6 (see [t
 If you're seeing those crashes, you could try setting `android.bundle.enableUncompressedNativeLibs=false` in your `gradle.properties`
 file. Be aware that this increases the size of your application when installed.
 
-Alternatively, you can use the `applyWorkaroundToOpenSqlite3OnOldAndroidVersions` method from this library.
+Alternatively, you can use the `applyWorkaroundToOpenSqlCipherOnOldAndroidVersions` method from this library.
 It will try to open `sqlcipher` in Java, which seems to work more reliably. After the native library has been loaded from Java,
 we can open it in Dart too.
--- a/sqlcipher_flutter_libs/README.md
+++ b/sqlcipher_flutter_libs/README.md
@@ -10,7 +10,7 @@ If you're seeing those crashes, you could try setting `android.bundle.enableUnco
 file. Be aware that this increases the size of your application when installed.
 
 Alternatively, you can use the `applyWorkaroundToOpenSqlite3OnOldAndroidVersions` method from this library.
-It will try to open `sqlite3` in Java, which seems to work more reliably. After sqlite3 has been loaded from Java,
+It will try to open `sqlcipher` in Java, which seems to work more reliably. After the native library has been loaded from Java,
 we can open it in Dart too.
 
 ### working with Android
--- a/sqlcipher_flutter_libs/README.md
+++ b/sqlcipher_flutter_libs/README.md
@@ -3,16 +3,6 @@
 Flutter apps depending on this package will
 contain native `SQLCipher` libraries.
 
-### Problems on Android 6
-
-There appears to be a problem when loading native libraries on Android 6 (see [this issue](https://github.com/simolus3/moor/issues/895#issuecomment-720195005)).
-If you're seeing those crashes, you could try setting `android.bundle.enableUncompressedNativeLibs=false` in your `gradle.properties`
-file. Be aware that this increases the size of your application when installed.
-
-Alternatively, you can use the `applyWorkaroundToOpenSqlite3OnOldAndroidVersions` method from this library.
-It will try to open `sqlcipher` in Java, which seems to work more reliably. After the native library has been loaded from Java,
-we can open it in Dart too.
-
 ### working with Android
   on android you need to override the open method like this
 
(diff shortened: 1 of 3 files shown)


sqlcipher_flutter_libs/README.md
```diff
@@ -1,7 +1,27 @@
...
 
+### Problems on Android 6
+
+There appears to be a problem when loading native libraries on Android 6 (see [this issue](https://github.com/simolus3/moor/issues/895#issuecomment-720195005)).
+If you're seeing those crashes, you could try setting `android.bundle.enableUncompressedNativeLibs=false` in your `gradle.properties`
+file. Be aware that this increases the size of your application when installed.
+
+Alternatively, you can use the `applyWorkaroundToOpenSqlite3OnOldAndroidVersions` method from this library.
+It will try to open `sqlite3` in Java, which seems to work more reliably. After sqlite3 has been loaded from Java,
+we can open it in Dart too.
```
simolus3, line 14:
```suggestion
It will try to open `sqlcipher` in Java, which seems to work more reliably. After the native library has been loaded from Java,
we can open it in Dart too.
```

sqlcipher_flutter_libs/README.md
```diff
@@ -1,7 +1,27 @@
...
+
+### working with Android
+  on android you need to override the open method like this
+
+```dart
+  open.overrideFor(
+      OperatingSystem.android, openCipherOnAndroid);
+```
+
+### no change needed for ios
```
simolus3, line 24: This shouldn't be a heading, we can emphasize it by making it bold

```suggestion
__No changes are necessary for iOS__
```

sqlcipher_flutter_libs/lib/sqlcipher_flutter_libs.dart
```diff
@@ -0,0 +1,56 @@
...
+
+import 'dart:ffi';
+import 'dart:io';
+import 'dart:math';
+
+import 'package:flutter/services.dart';
+
+const _platform = MethodChannel('sqlcipher_flutter_libs');
+
+/// Workaround to open sqlite3 on old Android versions.
```
simolus3, line 16:
```suggestion
/// Workaround to open sqlcipher on old Android versions.
```

sqlcipher_flutter_libs/lib/sqlcipher_flutter_libs.dart
```diff
@@ -0,0 +1,56 @@
...
+///
+/// On old Android versions, this method can help if you're having issues
+/// opening sqlite3 (e.g. if you're seeing crashes about `libsqlcipher.so` not
+/// being available). To be safe, call this method before using apis from
+/// `package:sqlite3` or `package:moor/ffi.dart`.
+///
+/// Big thanks to [@knaeckeKami](https://github.com/knaeckeKami) for finding
+/// this workaround!!
+Future<void> applyWorkaroundToOpenSqlite3OnOldAndroidVersions() async {
```
simolus3, line 25: We might want to call this `applyWorkaroundToOpenSqlCipherOnOldAndroidVersions`, but either way should be fine.

sqlcipher_flutter_libs/README.md
```diff
@@ -1,7 +1,27 @@
...
+It will try to open `sqlite3` in Java, which seems to work more reliably. After sqlite3 has been loaded from Java,
+we can open it in Dart too.
+
+### working with Android
+  on android you need to override the open method like this
+
+```dart
+  open.overrideFor(
+      OperatingSystem.android, openCipherOnAndroid);
+```
```
simolus3, line 22: This doesn't mention the workaround, so maybe we should move the general section about Android above the note on Android 6.0.1 to make that clearer?

simolus3 (commented):
Thanks :+1: I have some comments, mainly regarding the documentation.

sqlcipher_flutter_libs/lib/sqlcipher_flutter_libs.dart
Ali1Ammar: i didnt call it  because every programmer may have a diffrent usecase to call it

in main.dart or in splash screen or before using database ...

sqlcipher_flutter_libs/README.md
Ali1Ammar: done

sqlcipher_flutter_libs/lib/sqlcipher_flutter_libs.dart
simolus3: Sorry, I meant that we might want to rename this method to `applyWorkaroundToOpenSqlCipherOnOldAndroidVersions`, since the workaround we're applying is for sqlcipher instead of sqlite3.

Ali1Ammar: ohh sorry about that i rename it

simolus3:
Thanks :+1:

PR #35, opened 2021-05-20
Upgrade iOS sqlite pod to 3.35.4

kuhnroyal:
This was missed in the last update it seems.


simolus3:
Thanks! The analysis warnings are due to an `extra_pedantic` upgrade, I'll deal with that separately.

kuhnroyal:
Can you release this right away, kind of need this for the `RETURNING *` :)

simolus3:
I've just published this. Sorry for the delay, I was trying to upgrade everything to 3.35.5 while we're at it but it looks like syncing to Maven Central has some delays right now.

Issue #36, opened 2021-05-25
use Uint8List instead of list<int>


gcxfd:
see https://medium.com/flutter-community/working-with-bytes-in-dart-6ece83455721

simolus3:
Can you please be more specific on what the issue is? This package uses `Uint8List` internally all the time, and sqlite blobs are returned as `Uint8List` as well.

gcxfd:
see the image below , i means function return Uint8List instead of list<int>
![image](https://user-images.githubusercontent.com/69608605/119584986-5aefb800-bdfc-11eb-9cd5-101392d9e7b6.png)

Leptopoda:
Hi @simolus3 what is the status here?
Are you open for PRs?

simolus3:
I still don't really see the problem here.

Whenever this package reports data from sqlite3, say because it's reading a `BLOB` column or because a `BLOB` value is passed to a user-defined function, it is already exporting the data as `Uint8List`.
For data that is passed from Dart to sqlite3, we accept `List<int>`'s in general. As `Uint8List` implements `List<int>`, you can already use `Uint8List`s to bind blob values to prepared statements. There way we allocate a `List<int>` as a native pointer is this:

https://github.com/simolus3/sqlite3.dart/blob/2ffd871606e0f43f68f1364d8f4e1428dcbf3792/sqlite3/lib/src/ffi/memory.dart#L34-L42

`setAll` will check if the argument passed is a `Uint8List` and does a memcpy in that case, so I expect no overhead of us accepting a general `List<int>`. The other way around is much more important, and we're consistently using `Uint8List`s there. Again, you can also pass `Uint8List`s to this package without issue.

-------

In the future, there might be a way in Dart to pass `Uint8List`s to C without copying - then there's a benefit of only accepting them. But we can still do an `is` check to determine whether to share the buffer or whether it needs to be copied.

Leptopoda:
Thank you for the fast response.
I was able to pin the issue to a faulty conversion within our code. I must have gotten a bit confused when initially debugging our issue.

Sorry for the noise

PR #37, opened 2021-05-26
use Uint8List instead of List<int>

gcxfd:
https://github.com/simolus3/sqlite3.dart/issues/36

### DIFF
--- a/sqlite3/lib/src/api/functions.dart
+++ b/sqlite3/lib/src/api/functions.dart
@@ -3,7 +3,7 @@ import 'package:meta/meta.dart';
 /// A scalar function exposed to sql.
 ///
 /// {@template sqlite3_function_behavior}
-/// The function must either return a `bool`, `num`, `String`, `List<int>` or
+/// The function must either return a `bool`, `num`, `String`, `Uint8List` or
 /// `null`.
 ///
 /// If invoking the function throws a Dart exception, the sql function will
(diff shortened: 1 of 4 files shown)


sqlite3/lib/src/ffi/ffi.dart
```diff
@@ -69,6 +69,8 @@ extension ValueUtils on Pointer<sqlite3_value> {
   }
 }
 
+final utf8Encode = utf8.encoder.convert;
```
simolus3, line 72: I think we should only define this once, the usage in `memory.dart` can just import this one.

simolus3 (commented):
Thanks! I think we should also change `allocateBytes` to only accept an `Uint8List`. Currently it copies the data it gets to a new `Uint8List`, we can avoid that copy if we get an `Uint8List` in the first place.

sqlite3/lib/src/api/functions.dart
```diff
@@ -3,7 +3,7 @@ import 'package:meta/meta.dart';
 /// A scalar function exposed to sql.
 ///
 /// {@template sqlite3_function_behavior}
-/// The function must either return a `bool`, `num`, `String`, `List<int>` or
+/// The function must either return a `bool`, `num`, `String`, `Uint8List` or
```
simolus3, line 6: I don't think we should force users to return a `Uint8List` here. It's a breaking change which IMO is not necessary. Most users return a `Uint8List` already, I don't see much value in an enforcement in the type system.

We can check if we got an `Uint8List` and implicitly copy to an `Uint8List` if we didn't.

PR #38, opened 2021-05-28
MacOS support and SQLCipher update to 4.4.3

MobiliteDev:
Hi,

I've added the MacOS support for SQLCipher.
Tested on MacMini in a flutter app, with macOS 11.4.
I've updated the dependency SQLCipher from 4.4.0 to 4.4.3.

The SQLCipher readme is updated too, because on MacOS, you need to use a custom open function 'openCipherOnMacOS'


sqlcipher_flutter_libs/lib/sqlcipher_flutter_libs.dart
```diff
@@ -55,3 +55,7 @@ DynamicLibrary openCipherOnAndroid() {
     return DynamicLibrary.open('/data/data/$appId/lib/libsqlcipher.so');
   }
 }
+
+DynamicLibrary openCipherOnMacOS() {
+  return DynamicLibrary.process();
```
simolus3, line 60: I wonder if we should move this into the `sqlite3` package.

In the opening code on macOS, I think we could check if `DynamicLibrary.process()` contains sqlite3 symbols (e.g. with `lookup('sqlite3_version').address != 0`). If it does, we use `DynamicLibrary.process()`. If it doesn't, we fallback to the system library.

MobiliteDev:
I removed openCipherOnMacOS function en modified in sqlite3 like this, is it good for you ?
image

simolus3:
Thanks for your contribution :+1:

Issue #41, opened 2021-06-07
Add constants


MobiliteDev:
I found this example on the official dart repo, an FFI implementation of SQLite.
https://github.com/dart-lang/sdk/tree/master/samples/ffi/sqlite

In this example, the class `contants.dart` [defines](https://github.com/dart-lang/sdk/blob/master/samples/ffi/sqlite/lib/src/bindings/constants.dart) more SQLite constants knew [here](https://www.sqlite.org/rescode.html#done).

What do you think of this approach ? 
Do you think it could be a good idea to update sqlite3 to looks like to it ?

If yes, I could help with a PR.

simolus3:
The constants defined in this library are all internal and not part of any public api.

I think we can add error codes to the public api since they're generally helpful to help distinguish between exceptions. A PR would be appreciated, thanks!

PR #42, opened 2021-06-08
Add constants (#41)

MobiliteDev:
I've added all the constants, renamed the class name with "Sql" prefix.
Added to the sqlite3.dart for public visibility

### DIFF
--- a/sqlite3/lib/src/ffi/constants.dart
+++ b/sqlite3/lib/src/ffi/constants.dart
@@ -9,8 +9,8 @@ import 'dart:ffi';
 /// New error codes may be added in future versions of SQLite.
 ///
 /// See also: SQLITE_IOERR_READ | extended result codes,
-/// sqlite3_vtab_on_conflict() SQLITE_ROLLBACK | result codes.
-class SqlErrors {
+/// sqlite3_vtab_on_conflictstatic const int ) SQLITE_ROLLBACK | result codes.
+class SqlError {
   /// Successful result
   static const int SQLITE_OK = 0;
 
@@ -32,13 +32,13 @@ class SqlErrors {
   /// A table in the database is locked
   static const int SQLITE_LOCKED = 6;
 
-  /// A malloc() failed
+  /// A mallocstatic const int ) failed
   static const int SQLITE_NOMEM = 7;
 
   /// Attempt to write a readonly database
   static const int SQLITE_READONLY = 8;
 
-  /// Operation terminated by sqlite3_interrupt()
+  /// Operation terminated by sqlite3_interruptstatic const int )
   static const int SQLITE_INTERRUPT = 9;
 
   /// Some kind of disk I/O error occurred
@@ -47,7 +47,7 @@ class SqlErrors {
   /// The database disk image is malformed
   static const int SQLITE_CORRUPT = 11;
 
-  /// Unknown opcode in sqlite3_file_control()
+  /// Unknown opcode in sqlite3_file_controlstatic const int )
   static const int SQLITE_NOTFOUND = 12;
 
   /// Insertion failed because database is full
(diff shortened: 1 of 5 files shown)


MobiliteDev:
@simolus3 these CI errors are on your side I think.

image

sqlite3/lib/src/ffi/constants.dart
```diff
@@ -1,44 +1,247 @@
...
+  static const int SQLITE_DONE = 101;
+}
+
+/// Flags for file open operations, https://www.sqlite.org/c3ref/c_open_autoproxy.html
+/// Flags For File Open Operations
+///
+/// These bit values are intended for use in the
+/// 3rd parameter to the [sqlite3_open_v2()] interface and
+/// in the 4th parameter to the [sqlite3_vfs.xOpen] method.
+class SqlFlags {
```
simolus3, line 114: Do we need to add those flags to the public api? I think we don't provide an api where users could pass one of them, so I think we shouldn't make them public.

sqlite3/lib/src/ffi/constants.dart
```diff
@@ -1,44 +1,247 @@
...
+  static const int SQLITE_OPEN_PRIVATECACHE = 0x00040000;
+
+  /// VFS only
+  static const int SQLITE_OPEN_WAL = 0x00080000;
+}
 
 // Prepare flags, https://www.sqlite.org/c3ref/c_prepare_normalize.html
-const SQLITE_PREPARE_PERSISTENT = 0x01;
-const SQLITE_PREPARE_NO_VTAB = 0x04;
+class SqlPrepareFlags {
```
simolus3, line 177: Same here I think?

sqlite3/lib/src/ffi/constants.dart
```diff
@@ -1,44 +1,247 @@
...
+  ///The SQLITE_PREPARE_NORMALIZE flag is a no-op. This flag used to be required for any prepared statement that wanted to use the sqlite3_normalized_sql() interface.
+  /// However, the sqlite3_normalized_sql() interface is now available to all prepared statements, regardless of whether or not they use this flag.
+  static const int SQLITE_PREPARE_NORMALIZE = 0x02;
+
+  ///the SQL compiler to return an error (error code SQLITE_ERROR) if the statement uses any virtual tables
+  static const int SQLITE_PREPARE_NO_VTAB = 0x04;
+}
+
+/// Datatypes, https://sqlite.org/c3ref/c_blob.html
+class SqlTypes {
```
simolus3, line 193: This one is exposed through Dart types, users shouldn't see those.

sqlite3/lib/src/ffi/constants.dart
```diff
@@ -1,44 +1,247 @@
...
+  static const int SQLITE_UTF16_ALIGNED = 8;
+}
 
-const SQLITE_INTEGER = 1;
-const SQLITE_FLOAT = 2;
-const SQLITE_TEXT = 3;
-const SQLITE_BLOB = 4;
-const SQLITE_NULL = 5;
+/// Special destructors, https://www.sqlite.org/c3ref/c_static.html
+class SqlSpecialDestructor {
```
simolus3, line 224: Nice call to introduce fields for these values! Let's just not expose them to users.

sqlite3/lib/src/ffi/constants.dart
```diff
@@ -1,44 +1,247 @@
...
+/// Result Codes
+///
+/// Many SQLite functions return an integer result code from the set shown
+/// here in order to indicates success or failure.
+///
+/// New error codes may be added in future versions of SQLite.
+///
+/// See also: SQLITE_IOERR_READ | extended result codes,
+/// sqlite3_vtab_on_conflict() SQLITE_ROLLBACK | result codes.
+class SqlErrors {
```
simolus3, line 13: Let's call this class `SqlError` which makes it easier to read a specific constant in code.

simolus3 (commented):
My server should be fixed now, sorry for that.

I think we should not expose all of these constants since most of them aren't actually useful for users. You can add a `show SqlError` to only expose those.

sqlite3/lib/src/ffi/constants.dart
MobiliteDev: You are right, I've changed the visibility in sqlite3.dart on the last commit

sqlite3/lib/src/ffi/constants.dart
MobiliteDev: You are right, I've changed the visibility in sqlite3.dart on the last commit

sqlite3/lib/src/ffi/constants.dart
MobiliteDev: You are right, I've changed the visibility in sqlite3.dart on the last commit

sqlite3/lib/src/ffi/constants.dart
MobiliteDev: You are right, I've changed the visibility in sqlite3.dart on the last commit

sqlite3/lib/src/ffi/constants.dart
MobiliteDev: You are right, I've renamed to SqlError

MobiliteDev:
I've removed the 's' from all my classes.

and I've added SqlExtError in public visibility, you get it with `sqlite3_extended_result_codes` function in `database.dart`

simolus3:
I think the CI failures are due to the Flutter dev SDK not having the necessary Dart version yet. Thanks for your contribution, this looks good to me :+1:

PR #46, opened 2021-07-13
feat: add last SQL command to SqliteException

MobiliteDev:
Hi,

I've added the last SQL command on SQLiteException in order to have more details on the error.

This will give an error like this
image

### DIFF
--- a/sqlite3/lib/src/api/exception.dart
+++ b/sqlite3/lib/src/api/exception.dart
@@ -21,14 +21,24 @@ class SqliteException implements Exception {
   /// code, providing some idea of the cause of the failure.
   int get resultCode => extendedResultCode & 0xFF;
 
-  SqliteException(this.extendedResultCode, this.message, [this.explanation]);
+  /// Last SQL command executed
+  String sqlCommand;
+
+  SqliteException(this.extendedResultCode, this.message, this.sqlCommand,
+      [this.explanation]);
 
   @override
   String toString() {
     if (explanation == null) {
-      return 'SqliteException($extendedResultCode): $message';
+      return '''
+SqliteException($extendedResultCode): $message
+    SQL: $sqlCommand
+      ''';
     } else {
-      return 'SqliteException($extendedResultCode): $message, $explanation';
+      return '''
+SqliteException($extendedResultCode): $message, $explanation
+    SQL: $sqlCommand
+''';
     }
   }
 }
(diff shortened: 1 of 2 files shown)


sqlite3/lib/src/api/exception.dart
```diff
@@ -21,14 +21,24 @@ class SqliteException implements Exception {
   /// code, providing some idea of the cause of the failure.
   int get resultCode => extendedResultCode & 0xFF;
 
-  SqliteException(this.extendedResultCode, this.message, [this.explanation]);
+  /// Last SQL command executed
+  String sqlCommand;
+
+  SqliteException(this.extendedResultCode, this.message, this.sqlCommand,
+      [this.explanation]);
```
simolus3, line 28: Changing the constructor like this is a breaking change. Can we make `sqlCommand` nullable and then use this instead?

```suggestion
  SqliteException(this.extendedResultCode, this.message,
      [this.explanation, this.sqlCommand]);
```

sqlite3/lib/src/api/exception.dart
```diff
@@ -21,14 +21,24 @@ class SqliteException implements Exception {
   /// code, providing some idea of the cause of the failure.
   int get resultCode => extendedResultCode & 0xFF;
 
-  SqliteException(this.extendedResultCode, this.message, [this.explanation]);
+  /// Last SQL command executed
+  String sqlCommand;
```
simolus3, line 25: This should be final, I also think we should use "statement" instead of "command":

```suggestion
  /// The SQL statement triggering this exception.
  final String? sqlStatement;
```

simolus3 (commented):
Thanks! I think this can be very helpful.

sqlite3/lib/src/api/exception.dart
MobiliteDev: Done

simolus3:
I think we should also add this to other potential error sources like prepared statements. That will be more complicated though. I can do it later today if you don't want to.

MobiliteDev:
> I think we should also add this to other potential error sources like prepared statements. That will be more complicated though. I can do it later today if you don't want to.

You are right.
But honestly, I don't feel able to edit your sources yet, I tried, but I think it's too complicated for me now.
If you're still OK, I prefer you to edit theses sources.

simolus3:
I've squashed and merged this with 26df1fceacbf71e49457c8021e925d1791a5d75d - thanks a lot for your contribution!

PR #53, opened 2021-10-17
Extract the column's table name in select results

juancastillo0:
Hi, thanks for your work.

This creates a new function in `Row`, `toTableColumnMap` similar to the one implemented in the [postgres](https://github.com/isoos/postgresql-dart/blob/b43e6ac8de66ed1e0b8efe52182cff1d81bc104b/lib/src/connection.dart#L571) package. It uses an empty string as a replacement for the null value returned from sqlite. Maybe it would be better to return a `Map<String?, Map<String, dynamic>>`, but I implemented it this way because that's the API in postgres.
Also, I created a helper function in `readNullableString` in the `Utf8Utils` extension which checks for a null pointer before calling `readString`.
There is also a test with an inner join select * and a computed column.

Please, tell me if there is anything else I can help you with. I can make the table name nullability change if you think that's a better API. Thanks.

### DIFF
--- a/sqlite3/lib/src/impl/statement.dart
+++ b/sqlite3/lib/src/impl/statement.dart
@@ -61,10 +61,10 @@ class PreparedStatementImpl implements PreparedStatement {
       return null;
     }
     final columnCount = _bindings.sqlite3_column_count(_stmt);
-    return Iterable.generate(columnCount, (i) {
+    return List.generate(columnCount, (i) {
       final pointer = _bindings.sqlite3_column_table_name!(_stmt, i);
       return pointer.isNullPointer ? null : pointer.readString();
-    }).toList();
+    });
   }
 
   @override
--- a/sqlite3/lib/src/impl/statement.dart
+++ b/sqlite3/lib/src/impl/statement.dart
@@ -55,6 +55,16 @@ class PreparedStatementImpl implements PreparedStatement {
     ];
   }
 
+  List<String> get _tableNames {
+    final columnCount = _bindings.sqlite3_column_count(_stmt);
+
+    return Iterable.generate(columnCount, (i) {
+      final _tableName =
+          _bindings.sqlite3_column_table_name(_stmt, i).readNullableString();
+      return _tableName ?? '';
+    }).toList();
+  }
+
   @override
   ResultSet select([List<Object?> parameters = const <Object?>[]]) {
     _ensureNotFinalized();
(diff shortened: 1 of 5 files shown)


simolus3:
Thanks for the PR! I'm slightly leaning towards using `null` for columns that are not part of any table (both in `toTableColumnMap` and in `tableNames`), but either way is fine for me.

Also, it looks like the native function we're using is only available when sqlite3 is compiled with `DSQLITE_ENABLE_COLUMN_METADATA`. I think we shouldn't assume this function to be available unconditionally.

You can use [`providesSymbol`](https://api.dart.dev/stable/2.14.4/dart-ffi/DynamicLibrary/providesSymbol.html) to check if a function is available in a library. If we don't have this function, maybe we could make `tableNames` and `toTableColumnMap` nullable, saying that they're not available on all platforms.

juancastillo0:
Great, thanks for the guidance.
The last commit adds those changes, it makes `tableNames` and `toTableColumnMap` nullable with some documentation.
Also, the tables names can be null, the return type is now `Map<String?, Map<String, dynamic>>`. 
Reading some more parts of the codebase I deleted the `readNullableString` function, now it just checks if it's a null pointer and returns null inline. 
The `Bindings.sqlite3_column_table_name` function is null when `providesSymbol` returns `false`.
The test now uses an aliased column showing that you still have to be careful when joining with the same table, since the result would be merged for the same table name. The table name returned by sqlite is un-aliased.
Thanks

sqlite3/lib/src/impl/statement.dart
```diff
@@ -55,6 +55,18 @@ class PreparedStatementImpl implements PreparedStatement {
     ];
   }
 
+  List<String?>? get _tableNames {
+    if (_bindings.sqlite3_column_table_name == null) {
+      // unsupported
+      return null;
+    }
+    final columnCount = _bindings.sqlite3_column_count(_stmt);
+    return Iterable.generate(columnCount, (i) {
```
simolus3, line 64: Small nit: Let's just use `List.generate` directly instead of calling `toList()` in the end.

sqlite3/test/database_test.dart
```diff
@@ -82,6 +82,51 @@ void main() {
               'INSERT INTO foo VALUES (?); INSERT INTO foo VALUES (?);', [123]),
           throwsArgumentError);
     });
+
+    test('inner join with toTableColumnMap and computed column', () {
```
simolus3, line 86: I think we would have to skip this test if `sqlite3_column_table_name` is not available.

You can do something like

```dart
final hasColumnMeta = open.openSqlite().providesSymbol('sqlite3_column_table_name');

test('inner join with toTableColumnMap and computed column', () {
  // ...
}, skip: hasColumnMeta ? null : 'sqlite3 was compiled without column metadata');
```

Issue #63, opened 2021-12-08
named parameters


tofutim:
Is it possible to used named parameters in PreparedStatement? For example, in my old code, I have "WHERE Entries MATCH @query" and in C# before I would use bind("@query", "stuff"), and then execute.

https://sqlite.org/c3ref/bind_blob.html

My old code had:
```
				cmdAttachDatabase.Bind("@dbname", DbName);
                cmdAttachDatabase.Bind("@dbfullpath", _initDbPath);     // This may contain the CEROD password
				SQLite3.Result result = (SQLite3.Result)cmdAttachDatabase.ExecuteNonQuery();
				if (result != SQLite3.Result.OK)
				{
					throw new Exception("Unable to attach database");
				}
```
with
```
        internal static string AttachDatabase
        {
            get
            {
                return @"ATTACH @dbfullpath AS @dbname;";
            }
        }
```

https://sqlite.org/c3ref/bind_parameter_index.html

An `execute` and `select` that takes `Map<String, Object?>` might do the trick.

tofutim:
I haz the codez.

PR #65, opened 2021-12-09
add execute/select via map

tofutim:
This needs more testing, but conceptual PR I suppose for #63

### DIFF
--- a/sqlite3/assets/sqlite3.h
+++ b/sqlite3/assets/sqlite3.h
@@ -39,6 +39,9 @@ int sqlite3_reset(sqlite3_stmt *pStmt);
 
 int sqlite3_column_count(sqlite3_stmt *pStmt);
 int sqlite3_bind_parameter_count(sqlite3_stmt *pStmt);
+sqlite3_char *sqlite3_bind_parameter_name(sqlite3_stmt*, int);
+int sqlite3_bind_parameter_index(sqlite3_stmt*, sqlite3_char *zName);
+int sqlite3_clear_bindings(sqlite3_stmt*);
 sqlite3_char *sqlite3_column_name(sqlite3_stmt *pStmt, int N);
 
 int sqlite3_bind_blob64(sqlite3_stmt *pStmt, int index, void *data,
--- a/sqlite3/assets/sqlite3.h
+++ b/sqlite3/assets/sqlite3.h
@@ -41,7 +41,6 @@ int sqlite3_column_count(sqlite3_stmt *pStmt);
 int sqlite3_bind_parameter_count(sqlite3_stmt *pStmt);
 sqlite3_char *sqlite3_bind_parameter_name(sqlite3_stmt*, int);
 int sqlite3_bind_parameter_index(sqlite3_stmt*, sqlite3_char *zName);
-int sqlite3_clear_bindings(sqlite3_stmt*);
 sqlite3_char *sqlite3_column_name(sqlite3_stmt *pStmt, int N);
 
 int sqlite3_bind_blob64(sqlite3_stmt *pStmt, int index, void *data,
--- a/sqlite3/assets/sqlite3.h
+++ b/sqlite3/assets/sqlite3.h
@@ -39,7 +39,7 @@ int sqlite3_reset(sqlite3_stmt *pStmt);
 
 int sqlite3_column_count(sqlite3_stmt *pStmt);
 int sqlite3_bind_parameter_count(sqlite3_stmt *pStmt);
-int sqlite3_bind_parameter_index(sqlite3_stmt*, sqlite3_char *zName);
+int sqlite3_bind_parameter_index(sqlite3_stmt *, sqlite3_char *zName);
 sqlite3_char *sqlite3_column_name(sqlite3_stmt *pStmt, int N);
 
 int sqlite3_bind_blob64(sqlite3_stmt *pStmt, int index, void *data,
--- a/sqlite3/assets/sqlite3.h
+++ b/sqlite3/assets/sqlite3.h
@@ -39,7 +39,6 @@ int sqlite3_reset(sqlite3_stmt *pStmt);
 
 int sqlite3_column_count(sqlite3_stmt *pStmt);
 int sqlite3_bind_parameter_count(sqlite3_stmt *pStmt);
-sqlite3_char *sqlite3_bind_parameter_name(sqlite3_stmt*, int);
 int sqlite3_bind_parameter_index(sqlite3_stmt*, sqlite3_char *zName);
 sqlite3_char *sqlite3_column_name(sqlite3_stmt *pStmt, int N);
 
--- a/sqlite3/lib/src/api/statement.dart
+++ b/sqlite3/lib/src/api/statement.dart
@@ -26,7 +26,7 @@ abstract class PreparedStatement {
   /// [SqliteException] will be thrown.
   void execute([List<Object?> parameters = const <Object>[]]);
 
-  void executeMap([Map<String, Object?> parameters = const {}]);
+  void executeMap(Map<String, Object?> parameters);
 
   /// Selects all rows into a [ResultSet].
   ///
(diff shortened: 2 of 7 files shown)


sqlite3/lib/src/impl/statement.dart
```diff
@@ -220,6 +272,16 @@ class PreparedStatementImpl implements PreparedStatement {
           parameters, 'parameters', 'Expected $count parameters, got $length');
     }
   }
+
+  void _ensureMatchingMapParameters(Map<String, Object?>? parameters) {
+    final length = parameters?.length ?? 0;
+    final count = parameterCount;
+
+    if (length != count) {
```
simolus3, line 280: I think this check is not enough, I think we should also check that

1. For each key in the map, a matching SQL variable exists.
2. Each SQL variable is bound to a value from the Dart map.

If we check 1., checking 2. is just a matter of comparing parameter counts but we need to check 1. I think the most efficient way to check this would be to integrate this into `_bindMapParams`.

sqlite3/lib/src/ffi/sqlite3.g.dart
```diff
@@ -342,6 +342,42 @@ class Bindings {
   late final _sqlite3_bind_parameter_count = _sqlite3_bind_parameter_countPtr
       .asFunction<int Function(ffi.Pointer<sqlite3_stmt>)>();
 
+  ffi.Pointer<sqlite3_char> sqlite3_bind_parameter_name(
```
simolus3, line 345: This file is auto-generated, please don't change it manually. Instead, add

```C
const char *sqlite3_bind_parameter_name(sqlite3_stmt*, int);
```

to `assets/sqlite3.h` and then use `dart run ffigen` to re-generate this file.

sqlite3/lib/src/api/statement.dart
```diff
@@ -23,6 +23,8 @@ abstract class PreparedStatement {
   /// [SqliteException] will be thrown.
   void execute([List<Object?> parameters = const <Object>[]]);
 
+  void executeMap([Map<String, Object?> parameters = const {}]);
```
simolus3, line 26: I think we can make `executeMap` required here (just remove the default value). If a user explicitly types `.executeMap`, they can also provide the map.

```suggestion
  void executeMap(Map<String, Object?> parameters);
```

sqlite3/lib/src/api/statement.dart
```diff
@@ -32,6 +34,8 @@ abstract class PreparedStatement {
   /// [SqliteException] will be thrown.
   ResultSet select([List<Object?> parameters = const <Object>[]]);
 
+  ResultSet selectMap([Map<String, Object?> parameters = const {}]);
```
simolus3, line 37: Same here

```suggestion
  void selectMap(Map<String, Object?> parameters);
```

sqlite3/lib/src/impl/statement.dart
```diff
@@ -138,46 +168,68 @@ class PreparedStatementImpl implements PreparedStatement {
...
+    if (params == null || params.isEmpty) return;
+
+    // variables in sqlite are 1-indexed
+    for (var key in params.keys) {
+      final Object? param = params[key];
+
+      final keyBytes = utf8.encode(key);
+      final keyPtr = allocateBytes(keyBytes);
+      _allocatedWhileBinding.add(keyPtr);
+      final i = _bindings.sqlite3_bind_parameter_index(_stmt, keyPtr.cast());
```
simolus3, line 187: `i` could be zero here if no such parameter exists. I think we should throw an exception in that case.

simolus3 (commented):
Thanks for starting to work on this :+1:

tofutim:
@simolus3 is it possible to support bool in bindParam? I'll work further on this PR.
```
### ArgumentError: Invalid argument (params[9]): Allowed parameters must either be null or an int, num, String or List<int>.: false
```

simolus3:
> is it possible to support bool in bindParam?

So the thing is that there's no `sqlite3_bind_boolean`, but IMO having support for `bool` by mapping it to `sqlite3_bind_int64(value ? 1 : 0)` is reasonable. After all, the `TRUE` and `FALSE` keywords in sqlite3 are doing the same thing.

tofutim:
Something is wrong. sqlite3 inconsistently and erratically returns 0 for the index for bind_parameter.  Explains some of the issues I had which also seemed random. I'll have to investigate more. This is the index number returned, followed by the named parameter and the value. For three inserts. Notice that the 0's randomly appear!
```
0 - @fieldid c51c15b8-55ee-4aa6-928e-c35a62e30e86
2 - @name word
3 - @internalname word
0 - @datatype Text
5 - @displaytype Text
6 - @indextype None
7 - @ix -1
0 - @pos -1
9 - @iskey false
10 - @isreadonly true
0 - @isreadonlyreadonly false
12 - @ishidden false
0 - @issystem false


0 - @fieldid f81dace2-18d0-439f-9970-b6b6e0489ca6
0 - @name definition
0 - @internalname definition
4 - @datatype Text
5 - @displaytype Text
6 - @indextype FullText
7 - @ix -1
8 - @pos -1
9 - @iskey false
0 - @isreadonly true
11 - @isreadonlyreadonly false
0 - @ishidden false
0 - @issystem false


0 - @fieldid a3e7113e-7c4f-4c0e-960b-96380132dc9e
0 - @name entryorder
3 - @internalname entryorder
4 - @datatype Integer
5 - @displaytype Text
6 - @indextype None
7 - @ix -1
8 - @pos -1
9 - @iskey false
10 - @isreadonly true
0 - @isreadonlyreadonly false
0 - @ishidden true
13 - @issystem false
```

tofutim:
I checked this and was able to fix it with `final keyPtr = allocateBytes(keyBytes, additionalLength: 1);` i.e., adding a null byte at the end of the string. `sqlite3_bind_parameter_index` has no provision for length.

simolus3:
> Note: I don't really know now if additionalLength makes a null byte, do you know?*

Yes, `allocateBytes` will write zeroes for the additional length, so you don't have to ensure that manually.

sqlite3/assets/sqlite3.h
```diff
@@ -39,6 +39,9 @@ int sqlite3_reset(sqlite3_stmt *pStmt);
 
 int sqlite3_column_count(sqlite3_stmt *pStmt);
 int sqlite3_bind_parameter_count(sqlite3_stmt *pStmt);
+sqlite3_char *sqlite3_bind_parameter_name(sqlite3_stmt*, int);
+int sqlite3_bind_parameter_index(sqlite3_stmt*, sqlite3_char *zName);
+int sqlite3_clear_bindings(sqlite3_stmt*);
```
simolus3, line 44: I think we can remove this function, we call `sqlite3_reset` when a statement is ran multiple times.

sqlite3/lib/src/impl/statement.dart
```diff
@@ -138,46 +168,75 @@ class PreparedStatementImpl implements PreparedStatement {
...
-      } else {
+    _variablesBound = true;
+  }
+
+  void _bindMapParams(Map<String, Object?>? params) {
+    if (params == null || params.isEmpty) return;
+
+    // _bindings.sqlite3_clear_bindings(_stmt);
+
+    // variables in sqlite are 1-indexed
```
simolus3, line 182: This comment is not really helpful here :D Maybe move it down to the `i == 0` check to explain why that's an error?

sqlite3/lib/src/impl/statement.dart
```diff
@@ -138,46 +168,75 @@ class PreparedStatementImpl implements PreparedStatement {
...
-          'params[$i]',
-          'Allowed parameters must either be null or an int, num, String or '
-              'List<int>.',
-        );
+            params, 'params', 'Key $key was not found in statement.');
       }
+      _bindParam(param, i);
     }
 
     _variablesBound = true;
```
simolus3, line 197: I started the mistake with `_ensureMatchingParameters`, but I think it would be easier if this method just checks the expected length. Then we don't have to call `_ensureMatchingMapParameters` at all use-sites and just have this check when binding parameters.

We should do the same thing with `_ensureMatchingParameters`, but I can take care of that too.

simolus3 (commented):
Thanks for the updates!

Btw, I'm happy to help with tests and documentation if you want to give me push access to this branch on your repo. But if you want to do that yourself that's fine with me too :D

tofutim:
I think I have given you access.

simolus3:
Thanks!

PR #68, opened 2021-12-20
Migrate maven repository from jcenter to mavenCentral

kelBurke:


simolus3:
Thanks!

Did you run into any issues with `jcenter()` being there (apart from it being bad practice)?. We (hopefully) shouldn't depend on anything being resolved to jcenter. Just wondering how urgent it is to publish this fix to pub.. If it's breaking builds for you, probably very.

kelBurke:
We're migrating solely for good practice, there's no breaking builds

PR #71, opened 2022-01-07
Add function create collation

MobiliteDev:
We just added access to the function **sqlite3_create_collation_v2**, that we need for internal use.  
Feel free to comment and correct.

Thank you !


sqlite3/lib/src/api/database.dart
```diff
@@ -80,6 +80,30 @@ abstract class Database {
...
+  /// receives two string parameters (**A** & **B**). [function] will return 0
+  /// if **A** and **B**
+  /// are considered equals. A negative value is returned if **A** is less than **B**,
+  /// but a positive if **A** is greater than **B**.
+  ///
+  ///
+  ///
+  /// For more information, see https://www.sqlite.org/c3ref/create_collation.html.
+  void createCollation({
+    required String collateName,
```
simolus3, line 103: Thanks for the helpful docs on this function. Just one minor nit, I think we can call `collateName` just `name` since the rest should be clear from the context.

sqlite3/lib/src/impl/function_store.dart
```diff
@@ -71,9 +87,32 @@ class FunctionPointerAndData {
...
+  int sizeB,
+  Pointer<Void> textB,
+) {
+  final functionId = udp.address;
+  final target = functionStore.readCollating(functionId);
+
+  final Uint8List bytesA = textA.copyRange(sizeA);
+  final Uint8List bytesB = textB.copyRange(sizeB);
+  final String txtA = String.fromCharCodes(bytesA);
+  final String txtB = String.fromCharCodes(bytesB);
```
simolus3, line 107: I'm fairly sure `fromCharCodes` copies internally, so we don't need to copy the pointers into Dart's heap first.

Can we just use something like `String.fromCharCodes(textA.cast<Uint8>().asTypedList(sizeA))` and avoid the `copyRange` call?

sqlite3/lib/src/impl/database.dart
```diff
@@ -294,6 +294,29 @@ class DatabaseImpl implements Database {
...
+    required String collateName,
+    required CollatingFunction function,
+  }) {
+    final storedFunction = functionStore.registerCollating(function);
+    final namePtr = _functionName(collateName);
+
+    final result = _bindings.sqlite3_create_collation_v2(
+      _handle,
+      namePtr.cast(), // zFunctionName
+      _eTextRep(false, false),
```
simolus3, line 308: This is a bit confusing since ORing the flags for determinism/direct only doesn't appear to be supported for collations. Let's just use `SqlTextEncoding.SQLITE_UTF8` here:

```suggestion
      SqlTextEncoding.SQLITE_UTF8,
```

sqlite3/lib/src/impl/function_store.dart
simolus3: And actually, my understanding is that `textA` and `textB` contain the text as a utf8 bytes? In that case we can just use `textA.cast<sqlite3_char>().readString(sizeA)`. `String.fromCharCodes` would only be correct if the texts were encoded with utf16-BE (I think?).

Maybe we should add a text using higher unicode chars like emoji to make sure we're getting the encoding right here.

simolus3 (commented):
Thanks, this is looking really good! I have some minor comments, mostly regarding how we read text back from sqlite3 :D

sqlite3/lib/src/impl/database.dart
MobiliteDev: I'm ok with that.

For informations, the possible values are : 
- SQLITE_UTF8
- SQLITE_UTF16LE
- SQLITE_UTF16BE
- SQLITE_UTF16
- SQLITE_UTF16_ALIGNED

sqlite3/lib/src/impl/function_store.dart
MobiliteDev: It's pretty new for us to play with String by this way, we will change that.

``textA`` and ``textB`` contain text encoded in the encoding specified by the ``eTextRep`` argument.

As **sqlite3.dart** use only UTF8 for the moment, may we can simply keep reading the text as UTF8?

sqlite3/lib/src/api/database.dart
MobiliteDev: Seems legit !
We will apply that

sqlite3/lib/src/impl/function_store.dart
simolus3: Yup, reading as utf-8 sounds like the safest idea to me.

simolus3:
Looks good, thanks for your contribution!

MobiliteDev:
You're welcome, we are happy to contribute !

Sorry that you have to correct the **sqlite3.h**, we didn't understand the analyzer's issue. But now, we will know how to format correctly ;)

PR #74, opened 2022-01-14
Use package:path's join function for more robust path creation

bryanoltman:
Cleaning up after an oversight in https://github.com/simolus3/sqlite3.dart/pull/73


simolus3:
Oh right, good point. Thanks for the follow-up!

PR #75, opened 2022-01-23
Use ABI-specific integer types

simolus3:
This will require Dart 2.16, but we can start using ABI-specific integer types with `ffigen` version 5.


simolus3:
Sine the ABI-specific integers will be part of `dart:ffi` in Dart 2.17, this should wait a bit longer.

PR #80, opened 2022-02-28
Web support

simolus3:
Work-in-progress support for the `sqlite3` package to access sqlite3 with WebAssembly. Ideally, this should have all of the features already supported by the FFI implementation, with compatible common interfaces and without _any_ additional JavaScript sources (so emscripten is off the table).

Todos:

- Write a custom VFS invoking Dart helpers to manage a file system.
- Implement a virtual file system ontop of IndexedDB, similar to the approach chosen by absurd js
- Tests!!


westito:
This replaces current web implementation of Drift? 
This can handle DartJS 64bit problem and binding SQLite INTEGER to Dart's `BigInt` class? (instead of `int` that can't handle 64bit on web)

simolus3:
> This replaces current web implementation of Drift?

It will do that eventually, but for now I'll keep the sql.js based implementation in drift and introduce this under another import. They're not directly compatible with each other.

> This can handle DartJS 64bit problem and binding SQLite INTEGER to Dart's BigInt class? (instead of int that can't handle 64bit on web)

This implementation uses Dart's BigInts for sqlite3 ints, yes. Supporting that in drift too will be additional effort though.

majdi21:
@simolus3 I see this was merged to main. Is it "ready" for use? I understand there maybe be issues, but I wanted to get your opinion on the status of this work.

simolus3:
This is by no means production ready and [the docs](https://pub.dev/documentation/sqlite3/latest/sqlite3.wasm/sqlite3.wasm-library.html) clearly state that as well. The wasm port also makes no compatibility guarantees at the moment, so in that library there may be breaking changes between minor versions.

The ideas from https://github.com/simolus3/sqlite3.dart/pull/90 and future support for web workers for persistence will help stabilize this, but it might take a while to get that done.

That said, if persistence is not crucial to you (e.g. if you're mainly using sqlite3 to cache data), the implementation is pretty stable and ready to use.

simolus3:
Yes exactly. If you're using the wasm library with a `FileSystem.inMemory()`, you should be fine. There's the general risk of using something new, but we share lots of tests between the native and the WebAssembly implementation so I'd say it should work well in general.

jeannelim:
@simolus3 Does this new implementation support database encryption?

PR #83, opened 2022-04-02
Lower linux cmake minimum version

Depau:
The current stable Flutter SDK bundles CMake 3.10.2 and it doesn't look like this library is using any new CMake features.

```
Launching lib/main.dart on Linux in debug mode...
Building Linux application...
CMake Error at flutter/ephemeral/.plugin_symlinks/sqlite3_flutter_libs/linux/CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.14 or higher is required.  You are running version 3.10.2


Exception: Unable to generate build files
```

### DIFF
--- a/sqlite3_flutter_libs/linux/CMakeLists.txt
+++ b/sqlite3_flutter_libs/linux/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.10)
 set(PROJECT_NAME "sqlite3_flutter_libs")
 project(${PROJECT_NAME} LANGUAGES C CXX)
 


simolus3:
We use [`FetchContent_MakeAvailable`](https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_makeavailable) which the docs say has been added in 3.14. This was the main motivation for using 3.14 as a required lower version, I'm not sure if it's safe to accept older versions too?

Depau:
Whoops, I missed it. No, it's not safe to accept older versions.

simolus3:
Do you think there's anything we could do about this? I didn't know that Flutter bundles their own cmake, it's using the one from the system for me.

Depau:
I think it bundles CMake if you use the snap version, as recommended by the Flutter "getting started" tutorial for GNU/Linux.

An option could be to use `FetchContent_GetProperties` as described (and not recommended) in the overview https://cmake.org/cmake/help/latest/module/FetchContent.html#id2

I eventually pinned the previous version of this library for my project.

simolus3:
I don't know much about snaps, but judging from [this](https://github.com/canonical/flutter-snap/blob/main/snap/snapcraft.yaml#L39) it looks like they just depend on the `cmake` package from the distribution? Ubuntu 21.10 should ship cmake 3.18, so I'm still not sure where you got the 3.10 from.

PR #89, opened 2022-04-15
Add file listing to VFS

westito:
For listing databases (ex. to delete them)

### DIFF
--- a/sqlite3/lib/src/wasm/file_system.dart
+++ b/sqlite3/lib/src/wasm/file_system.dart
@@ -35,6 +35,9 @@ abstract class FileSystem {
   /// otherwise.
   void deleteFile(String path);
 
+  /// List all files
+  List<String> listFiles();
+
   /// Returns the size of a file at [path] if it exists.
   ///
   /// Otherwise throws a [FileSystemException].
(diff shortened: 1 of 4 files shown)


sqlite3/lib/src/wasm/file_system.dart
```diff
@@ -35,6 +35,9 @@ abstract class FileSystem {
   /// otherwise.
   void deleteFile(String path);
 
+  /// List all files
+  List<String> listFiles();
```
simolus3, line 39: Let's expand this a bit:

```suggestion
  /// List all files.
  ///
  /// Returns a list of paths for files that currently exist in this file system.
  List<String> listFiles();
```

simolus3 (commented):
Thanks for your contribution!

In b2aeb721f0a34c5ac7a51f89ce6ae7f24ac5efb5, I've added simple file system tests in `test/wasm/file_system_test.dart`. Could you rebase from there and add a simple test for `listFiles()` too? Something that initially matches an empty list, creates a bunch of files, checks again, deletes some files, checks again should be enough.

simolus3:
The failing integration tests are due to https://github.com/subosito/flutter-action/issues/146, I'll run them manually and release this tomorrow. Thanks!

PR #90, opened 2022-04-23
Add VFS 2.0

westito:
I created a new, improved files system. First I tried to fix the original, but after I while I realised creating a separate one will be better and more backward compatible. However, I added a path validator with exception to temp files.

When I first try WASM with IndexDBFileSystem, the data wasn't saved. There was no error, and I have no clue what's going on. After hours of investigation I found that the `persistanceRoot` is not prepended to path! I used `persistanceRoot` other than '/'. I thought provide 'test_database' in `WasmDatabase(path:)` will open database as `/db/test_database`. The second problem is IndexDBFileSystem not checks path validity. Stores file in memory, but writing to IndexDB silently fails. This applies for temp files too, without workaround. Temp file path always starts with `/tmp`. With custom `persistanceRoot` temp files will never be saved to IndexedDB.


sqlite3/lib/src/wasm/file_system.dart
```diff
@@ -4,11 +4,13 @@ import 'dart:indexed_db';
 import 'dart:math';
 import 'dart:typed_data';
 
-import 'package:path/path.dart' as p show url;
+import 'package:path/path.dart' as p;
```
simolus3, line 7: Why did this have to change? I think we should avoid accidentally using anything but the `url` path context.

sqlite3/lib/src/wasm/file_system.dart
```diff
@@ -35,9 +37,16 @@ abstract class FileSystem {
   /// otherwise.
   void deleteFile(String path);
 
-  /// List all files stored in this file system.
+  /// Lists all files stored in this file system.
+  @Deprecated('Use files() instead')
```
simolus3, line 41: I think the deprecation notice should briefly explain why there's a new method now.

sqlite3/lib/src/wasm/file_system.dart
```diff
@@ -63,23 +72,40 @@ abstract class FileSystem {
...
   @override
   String toString() {
-    return 'FileSystemException($errorCode)';
+    return 'FileSystemException: ($errorCode) $message';
   }
 }
 
+/// An exception thrown by a [FileSystem] implementation when try to access file
+/// outside of persistence root
+class FileSystemAccessException extends FileSystemException {
```
simolus3, line 88: IMO we don't need a separate class for this, we can just throw a `FileSystemException` with that error code and message directly. Or do you see a reason for users to explicitly catch this type?

sqlite3/lib/src/wasm/file_system.dart
```diff
@@ -35,9 +37,16 @@ abstract class FileSystem {
   /// otherwise.
   void deleteFile(String path);
 
-  /// List all files stored in this file system.
+  /// Lists all files stored in this file system.
+  @Deprecated('Use files() instead')
   List<String> listFiles();
 
+  /// Lists all files stored in this file system.
+  List<String> files();
```
simolus3, line 45: While we're at it, I think `List<String> get files` would be more typical for a Dart API :)

sqlite3/lib/src/wasm/file_system_v2.dart
```diff
@@ -0,0 +1,298 @@
...
+/// A file system storing whole files in an IndexedDB database.
+///
+/// As sqlite3's file system is synchronous and IndexedDB isn't, no guarantees
+/// on durability can be made. Instead, file changes are written at some point
+/// after the database is changed. However you can wait for changes manually
+/// with [flush]
+///
+/// In the future, we may want to store individual blocks instead.
+
+class IndexedDbFileSystemV2 implements FileSystem {
```
simolus3, line 12: I like the new APIs, but is there no way of getting them into the existing indexed db implementation?

(It's ok to have mild breaking changes, the wasm API is marked as experimental and there's a documentation comment explaining that `package:sqlite3/wasm.dart` is not currently subject to semantic versioning).

sqlite3/lib/src/wasm/js_interop.dart
```diff
@@ -34,6 +34,24 @@ extension JsContext on _JsContext {
   external IdbFactory? get indexedDB;
 }
 
+extension IdbFactoryExt on IdbFactory {
+  Future<List<DatabaseName>> databases() async {
+    final jsDatabases = await promiseToFutureAsMap(
+        callMethod<Object>(this, 'databases', const []));
```
simolus3, line 40: In general, I try to avoid `js_util`, can we use `@staticInterop` instead? `@staticInterop` isn't documented that well, I can also help with the migration if you want me to.

sqlite3/lib/src/wasm/js_interop.dart
```diff
@@ -34,6 +34,24 @@ extension JsContext on _JsContext {
...
+    }).toList();
+  }
+}
+
+class DatabaseName {
+  final String name;
+  final int version;
+
+  DatabaseName(this.name, this.version);
+}
```
simolus3, line 53: Would something like this work?

```suggestion
@JS()
@anonymous
class DatabaseName {
  external String get name;
  external int get version;
}
```

When we can just do `promiseToFuture<DatabaseName>()` and don't have to go over the map?

simolus3 (commented):
Awesome stuff, thanks for your contributions :+1: 

I'm wondering if we have to introduce a separate class for the v2 API, do you think there's a way to add these methods to the existing class?

sqlite3/lib/src/wasm/js_interop.dart
westito: I'm new in JS interop. I tried a solution similar to this, but didn't worked. I try again

sqlite3/lib/src/wasm/file_system_v2.dart
westito: It can be simply replaced with current implementation if a breaking change is not a problem. Deprecation messages can be omitted also.

sqlite3/lib/src/wasm/file_system.dart
westito: I thought it is a more verbose exception to path validation and different from "regular" IO exceptions. But of course I can remove it. By the way, if current implementation will be replaced, this wouldn't need anymore.

sqlite3/lib/src/wasm/file_system.dart
westito: In deed. I use other methods from it. Was simpler to remove it while develop. I'll fix

westito:
I made some changes to JS interop to make it simpler. But I can't skip mapping, because `promiseToFuture` can't handle `List<>` type argument, so `promiseToFuture<List<DatabaseName>>()` not working. 
What I found:
https://github.com/dart-lang/sdk/issues/39627#issuecomment-571297372

westito:
I tried to add a legacy option to VFS2 and leave `load()` as deprecated function. But beyond `persistanceRoot` prepending, VFS2 use path normalisation. A lot of `if (legacy)` statements will make the code trashy. I think better replace the original implementation without backward compatible. Or create a separate V2 class.

westito:
Can we enable WAL mode in SQLite? We could achieve a better performance on this VFS I think

westito:
How SQLite handles data corruption? When bytes write with VFS, IndexedDB may can't persist because works async. SQLite think it is written, database will be corrupted next load. Am I right? Can we handle this, or simply warn users do not rely on data saved to SQLite? I am using SQLite for cache only, so corruption not a problem. App re-downloads database in this case.

simolus3:
> Can we enable WAL mode in SQLite?

My idea was to keep the sqlite3 build small which is why I disabled it. I can enable it if it turns out to improve performance.

> How SQLite handles data corruption?

I don't think we can really handle this: Even though IndexedDB provides ACID guarantees, the VFS API from sqlite requires us to be synchronous and IndexedDB is async, so we can't rely on that. The assumption is that writes will probably go through because they're fast, but as you said this isn't a safe assumption to make.

If you need synchronization I'd probably recommend using shared workers.

We could call asynchronous APIs by running them in a web worker and using [atomics](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics) to have the main script wait for them synchronously. That sounds complicated though.

westito:
That's awesome idea. With web compiler, FVS can be compiled to a single js file and loaded as shared worker. This way, it isn't necessary to compile to js and load the whole sqlite (or Drift) as shared worker.

westito:
I'm working on a block based storage. Works like a charm. I still have to clean the code and write tests

Képernyőfotó 2022-04-25 - 13 31 59

simolus3:
No pressure, but how's the block storage coming along? I feel Iike that approach would scale much better for large databases, so please let me know if there's anything I can do to help!

westito:
Sorry, I didn't had much time to finish in this week. Tomorrow I'll try to.

simolus3:
No worries, thanks for the update!

westito:
Finally done. Works with current tests and in real application. Need some edge-cased tests, update comments. Journal is written very often. Should give it a try to use WAL mode. I'm thinking on a sync version with shared worker as I mentioned before.

westito:
Storing files in separate objectStores seemed much simpler. Writing or inserting blocks can be faster in dedicated objects.(much less key lookup and index rebuild) and modifying metadata on each write can doubles the write time. Because of async operation, this doesn’t seem like a big deal, but slower write speed is more likely to case database corruption. Write speed will also be important in the shared worker version because of synchronous writing. I think using single object store + metadata makes things more complicated without adding benefits. Creating new file could be faster, but it's not a point here I think. But this is your library, its up to you how things going :)

westito:
Can you compile a new sqlite.wasm? tried on Mac without success. I copied and committed the latest wasm to sqlite3/example/web/sqlite3.wasm for tests in this PR. I think this should leave there and updated when sqlite3 source changes. I can't run this PR now because develop merged and my wasm is old.

simolus3:
Of course, you can grab the latest sqlite3.wasm from https://storage.googleapis.com/simon-public-euw3/assets/sqlite3.wasm (the same file is used for tests with GitHub actions).

simolus3:
I've pushed some changes to only use two object stores (one for file names and one for blobs). I've also moved the block storage into the IndexedDB file system, IMO there's no big advantage of also having that in the in-memory file system. I've created the `AsynchronousIndexedDbFileSystem` class which is essentially an asynchronous version of the file system interface. The `IndexedDbFileSystem` now wraps an async FS. I hope that this distinction will make it easier to write a say `WorkerBasedFileSystem` in the future where a worker is using an async FS and uses atomics to synchronize calls.

I've removed some path normalization logic you added because I didn't fully understand it. I think that sqlite3 will call our `path_normalize` function before using the file system, so it shouldn't be necessary? Is this to make API misuse harder? Might be reasonable to have it, but then we should use it consistently across all file system implementations.

simolus3:
I'll merge this and get this released tomorrow. Thanks for your continued great work on this package :rocket:

westito:
Thank you for improve my code. Yes, it's unnecessary having blocks in-memory. It was easier for me to test if I modified the original code as little as possible. I see you wrote your own cursor handler. I did not use the original because it discarded the transaction. The original implementation is very buggy. I don't know if it's worth dealing with and contributing to DartSDK.
I had a more general VFS in mind, so this path normalization idea came up.

simolus3:
> The original implementation is very buggy.

Yeah I feel like exposing IndexedDB cursors as streams was an unfortunate design choice as pauses or not immediately reacting to new rows will close the transaction. So it's very easy to cause deadlocks with that API.

> I had a more general VFS in mind, so this path normalization idea came up.

I think I want the VFS API to mainly be used by sqlite3 internally at the moment, and for that it's preferable to avoid path normalization logic for every write. It's not exactly user friendly, I'll see if it makes sense to polish that API with path normalization and expose it more prominently.

westito:
There is an error when I try to run in my app. I can't understand exactly the problem. The getKey returns with this exception. Not immediately, after inserts. Maybe when transaction commits. 

Képernyőfotó 2022-05-04 - 10 33 16
Képernyőfotó 2022-05-04 - 10 33 08
Képernyőfotó 2022-05-04 - 10 32 57
Képernyőfotó 2022-05-04 - 10 27 27

simolus3:
Yeah I just spent some time debugging this too :D I think it's because we don't reset cached file ids after a file is deleted, and the journal file gets deleted and re-created a lot. I've fixed this locally, but I need to write some tests to ensure this is working as intended.

westito:
I can send you a sample project. It is working when Drift loaded as worker. But run in main application this exception comes up. It is working with the previous non-block VFS.

simolus3:
> It is working when Drift loaded as worker.

I think you're just not seeing the error messages in the main log then, for me a similar error showed up when inspecting the worker. Also, workers are completely broken because instantiating a `KeyRange` will internally check the `window` to see which property to use. This doesn't work in a worker of course, so I had to use `package:js` to create `KeyRange`s...

I believe I have fixed this in 94d96527c117fcbe613c78bae8943b5effef4b6a, can you try that state in your project and see if that fixes the errors for you as well?

PR #95, opened 2022-05-11
Add BigInt support to FFI

westito:
- Added support for BigInt types to FFI
- Added int type to WASM binding. (is it intentional? int is converting to double right now)

### DIFF
--- a/sqlite3/lib/src/wasm/function_store.dart
+++ b/sqlite3/lib/src/wasm/function_store.dart
@@ -151,6 +151,8 @@ class FunctionStore {
   void _contextSetResult(Pointer context, Object? result) {
     if (result == null) {
       _bindings.sqlite3_result_null(context);
+    } else if (result is int) {
+      _bindings.sqlite3_result_int64(context, BigInt.from(result));
     } else if (result is BigInt) {
       _bindings.sqlite3_result_int64(context, result);
     } else if (result is double) {
--- a/sqlite3/lib/src/wasm/statement.dart
+++ b/sqlite3/lib/src/wasm/statement.dart
@@ -161,6 +161,8 @@ class WasmStatement extends CommonPreparedStatement {
   void _bindParam(Object? param, int i) {
     if (param == null) {
       bindings.sqlite3_bind_null(statement, i);
+    } else if (param is int) {
+      bindings.sqlite3_bind_int64(statement, i, BigInt.from(param));
     } else if (param is BigInt) {
       bindings.sqlite3_bind_int64(statement, i, param);
     } else if (param is bool) {
--- a/sqlite3/CHANGELOG.md
+++ b/sqlite3/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.7.1-dev
+
+- Allow binding `BigInt`s to statements and functions. They must still be
+  representable as a 64-bit int, but this closes a compatibility gap between
+  the web and the native implementations.
+
 ## 1.7.0
 
 - Add support for application-defined window functions. To register a custom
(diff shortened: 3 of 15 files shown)


westito:
There is an interesting issue. I inserted >64bit values to database.

On WASM it is overflows without error:
[        ] 9223372036854775800
[        ] 9223372036854775801
[        ] 9223372036854775802
[        ] 9223372036854775803
[        ] 9223372036854775804
[        ] 9223372036854775805
[        ] 9223372036854775806
[        ] 9223372036854775807
[        ] -9223372036854775808
[        ] -9223372036854775807
[        ] -9223372036854775806
[        ] -9223372036854775805
[        ] -9223372036854775804
[        ] -9223372036854775803

On FFI it is capped without error (tested on macOS, iOS, Android)
[        ] flutter: 9223372036854775800
[        ] flutter: 9223372036854775801
[        ] flutter: 9223372036854775802
[        ] flutter: 9223372036854775803
[        ] flutter: 9223372036854775804
[        ] flutter: 9223372036854775805
[        ] flutter: 9223372036854775806
[        ] flutter: 9223372036854775807
[        ] flutter: 9223372036854775807
[        ] flutter: 9223372036854775807
[        ] flutter: 9223372036854775807
[        ] flutter: 9223372036854775807
[        ] flutter: 9223372036854775807

Because SQLite not handles this properly, I think must be a bound check at library level.

sqlite3/lib/src/wasm/function_store.dart
```diff
@@ -151,6 +151,8 @@ class FunctionStore {
   void _contextSetResult(Pointer context, Object? result) {
     if (result == null) {
       _bindings.sqlite3_result_null(context);
+    } else if (result is int) {
+      _bindings.sqlite3_result_int64(context, BigInt.from(result));
```
simolus3, line 155: I think this might not work? `int` and `double` are the same type on the web. So if I want to insert `3.5`, `result is int` would be true and we'd bind the wrong type.

sqlite3/lib/src/wasm/statement.dart
```diff
@@ -161,6 +161,8 @@ class WasmStatement extends CommonPreparedStatement {
   void _bindParam(Object? param, int i) {
     if (param == null) {
       bindings.sqlite3_bind_null(statement, i);
+    } else if (param is int) {
+      bindings.sqlite3_bind_int64(statement, i, BigInt.from(param));
```
simolus3, line 165: Same here, it would be good to test this and ensure doubles still work properly.

simolus3 (commented):
> There is an interesting issue. I inserted >64bit values to database.

The ABI only allows 64bit values, so I think if we allow the conversion from BigInt we should manually check that we don't overflow when converting to an int.

sqlite3/lib/src/wasm/function_store.dart
westito: Under the hood Dart simply check it is fraction or not. So this type check order is good. But vice versa, if you check int type variable with `is double` it will always be true. 

```JavaScript
core.int = class int extends core.num {
    static is(o) {
      return typeof o == "number" && Math.floor(o) == o;
    }
}
core.double = class double extends core.num {
    static is(o) {
      return typeof o == "number";
    }
}
```

simolus3: Oh nice I didn't know about that, I thought the types were identical. To ensure this is working, can we add a test to ensure that `SELECT typeof(?)` executed with an integer as a parameter evaluates to `integer` on all platforms? Could be in `test/common/database.dart`.

westito: I can't wait for for dart2wasm 😄 https://github.com/dart-lang/sdk/commit/6faa5f3bd00ad8cbc640b3fc80cf7466c002a7df

westito:
I did some investigation. On web Dart integer (and JS integer too) capped to max. On native it is overflows to negative. Exactly like in SQLite. There is no bound check. I think we not need to add bound check, it is up to user to check this. Or be kind and add check before insert? 😄

westito:
I added bound check for BigInt and some tests for it

simolus3:
Thanks for the PR!

PR #97, opened 2022-05-19
Add test for indexeddb speed

westito:
There is a (big) issue with async IndexedDB fs. After inserting a bunch of rows, the persistence take a lot of time. I put 5sec waiting for simulate browser close. When reopen, database will be corrupted. 
Right now, all writes goes in a queue even these writes overlap/overwrites each other. My idea is adding dirty flag for blocks and debounce persistence writes. This way only changed blocks are written at once after bunch of memory writes.

### DIFF
--- a/sqlite3/test/wasm/utils.dart
+++ b/sqlite3/test/wasm/utils.dart
@@ -2,7 +2,7 @@ import 'package:http/http.dart' as http;
 import 'package:sqlite3/wasm.dart';
 import 'package:test/scaffolding.dart';
 
-Future<WasmSqlite3> loadSqlite3() async {
+Future<WasmSqlite3> loadSqlite3([SqliteEnvironment? environment]) async {
   final channel = spawnHybridUri('/test/wasm/asset_server.dart');
   final port = await channel.stream.first as int;
 
@@ -16,5 +16,5 @@ Future<WasmSqlite3> loadSqlite3() async {
         'Could not load module (${response.statusCode} ${response.body})');
   }
 
-  return WasmSqlite3.load(response.bodyBytes);
+  return WasmSqlite3.load(response.bodyBytes, environment);
 }
--- a/sqlite3/test/wasm/wasm_file_system_test.dart
+++ b/sqlite3/test/wasm/wasm_file_system_test.dart
@@ -1,0 +1,61 @@
+@Tags(['wasm'])
+
+import 'package:sqlite3/wasm.dart';
+import 'package:test/test.dart';
+
+import 'utils.dart';
+
+Future<void> main() async {
+  group('indexed db', () {
+    test('with proper persistence', () async {
+      final fileSystem = await IndexedDbFileSystem.open(dbName: 'test');
+      final sqlite3 =
+          await loadSqlite3(SqliteEnvironment(fileSystem: fileSystem));
+      final database = sqlite3.open('test');
+      expect(database.userVersion, 0);
+      database.userVersion = 1;
+      expect(database.userVersion, 1);
+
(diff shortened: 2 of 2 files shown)


simolus3:
Ok, so I think the main optimization that really drives this forward is to eliminate patterns of "create file, write into that, delete file". This is what sqlite3 is doing with a journal for _every_ write to the database, and we open an IndexedDB transaction for each of that. So there are essentially two optimizations here:

- When we write to a file, find previously scheduled pending writes to the same file and combine the chunks.
- When we delete a file, delete previously scheduled (but not yet started) writes to the same file and the matching creation of this file, if there is any.

With these two optimizations, the amount of operations drops from around 3000 to exactly 2 (create test database file + write content) :D

westito:
My previous solution stored data in memory as blocks already and did only full-block writes. It was pretty fast. Stored 500 inserts in less than 3sec

simolus3:
Yeah I've implemented something similar in 6a343806e46497fbd4a0c173af351522b74f6cb7. Once a write is scheduled, we keep a mutable map of full blocks that need to updated. For further writes that happen before we start to work on the initial write, that map is updated to also include these subsequent writes. So we'll always write a full block, but we'll also reduce the total amount of writes (now, when there are lots of writes to the same block _synchronously_, we just need to do one async write in the end).

westito:
Sorry, accidentally I watch memory logs 😅 My solution was slow also. But this last commit seems very fast! Great work.

westito:
Can you make a journal-like meta data for files? I thinking on solution that can prevent data corruption (more like prevent inconsistency). 
1. File marked as dirty until flush() called manually. 
2. In `WasmSqlite3` flush() called after every insert/update/etc (or after a transaction on bulk writes).
3. When reload filesystem, dirty files will be ignored and deleted. 
This can solve two issues: Application can be sure database persisted (awaits flush() after write queries), and can be sure there won't be inconsistency (for ex. data sent to backend and app updating the status while browser/app exits/crash). 
If user writes queries properly (for ex. do bulk inserts in transaction) it won't be slow. Of course, this would be an option (`syncWrite: true`). I think this could be replace our worker-based synchronised filesystem idea. That would have been too slow and complicated anyway. 
PS.: This improved FS is insane! I can't even understand the algorithm. 😄

westito:
> eliminate patterns of "create file, write into that, delete file". This is what sqlite3 is doing with a journal for _every_ write to the database
>
Set `locking_mode=EXCLUSIVE;` only truncate journal, not deletes the file. 
As I mentioned before, I think persisting WAL mode sqlite would be much faster because blocks are not overwritten, only new blocks added (until a checkpoint, that happens later - or triggered manually).

simolus3:
Instead of marking files as dirty, we could just not write them to the IndexedDB at all. There could be an `autoWrite: false` parameter which disables async work startiing without anyone awaiting it. Then we'd only start a work iteration if you call `flush()` or `close()`.

Thanks to the new optimizations the worklist accumulated between many writes would probably still be manageable if `flush()` is only called sporadically - _and_ we eliminate inconsistencies! 

> PS.: This improved FS is insane! I can't even understand the algorithm. :smile:

Yeah there's lots of documentation I still need to add :D

> Set `locking_mode=EXCLUSIVE;` only truncate journal, not deletes the file.

Yeah that's probably another case to optimize (a truncate call could remove pending writes writing to higher offsets in the file).

> Because the main database file modified only on checkpointing and not on every update, data corruption is far less likely.

Would be interesting to see how much of a difference that still makes with the faster FS implementation dropping most unnecessary writes. My main motivation to exclude WAL was to make the compiled WASM module as small as possible, but it probably doesn't make that big of a difference.

westito:
Worth a try.

PR #98, opened 2022-05-21
Indexeddb improvements

westito:
I accidentally delete my branch. This is the continuation of previous. I added a wire to make FS flushing available to database delegate. Further implementation of this will be in a new Drift PR.

### DIFF
--- a/sqlite3/lib/src/common/database.dart
+++ b/sqlite3/lib/src/common/database.dart
@@ -173,6 +173,9 @@ abstract class CommonDatabase {
     bool directOnly = true,
   });
 
+  /// Waits for the data to be written to persistent storage
+  Future<void> flush();
+
   /// Closes this database and releases associated resources.
   void dispose();
 }
--- a/sqlite3/lib/src/wasm/database.dart
+++ b/sqlite3/lib/src/wasm/database.dart
@@ -344,6 +344,11 @@ class WasmDatabase extends CommonDatabase {
     );
   }
 
+  @override
+  Future<void> flush() async {
+    await bindings.persistentStorage?.flush();
+  }
+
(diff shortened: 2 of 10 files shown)


sqlite3/lib/src/common/database.dart
```diff
@@ -173,6 +173,9 @@ abstract class CommonDatabase {
     bool directOnly = true,
   });
 
+  /// Waits for the data to be written to persistent storage
+  Future<void> flush();
+
```
simolus3, line 178: This API definitely shouldn't be on the common database since it doesn't make sense for all implementations. If at all, we should only add it to the wasm database.

sqlite3/lib/src/wasm/database.dart
```diff
@@ -344,6 +344,11 @@ class WasmDatabase extends CommonDatabase {
     );
   }
 
+  @override
+  Future<void> flush() async {
+    await bindings.persistentStorage?.flush();
+  }
```
simolus3, line 350: It feels a bit weird to have an API for this on the database to be honest.

I rather change the drift wasm API to also manage the file system and call `flush()` directly instead of doing it through the database here.

simolus3 (commented):
I didn't take a look at the drift-side of this yet, but I wonder if we really need this api or if we can just ask drift to manage things.

Maybe we can add a `SqliteEnvironment get environment` getter to `WasmSqlite` and let drift use that instead of going through the database?

simolus3:
I've merged my IndexedDb changes, it turned out that the rest isn't needed for drift.

Other libraries can go the same way and manage the file system for users too :D Thanks for your work on this and the drift PR!

PR #115, opened 2022-08-23
Added support for spellfix1 module

FaFre:
This PR adds general support for loading statically linked extensions via `sqlite3_auto_extension` and is also adding the spellfix1 extension (https://www.sqlite.org/spellfix1.html).


FaFre:
@simolus3  can you please check, if you pushed the last changes for `sqlite-native-libraries`?

 The patch version should be `2` as referenced in https://github.com/simolus3/sqlite3.dart/blob/1621cce660ee4f64992910a20b5492112f675334/sqlite3_flutter_libs/android/build.gradle#L29 but it is still `0` (https://github.com/simolus3/sqlite-native-libraries/blob/a1e3bafbb2cd27d9a8a4526b09681ecd65da0d44/sqlite3-native-library/build.gradle.kts#L12). I remember when first compiling the android native libraries I had to increase this number. If this here is ok I will issue an PR there as well.

Is the extension loading itself is approved by you I will add some tests for this.

simolus3:
My bad on not publishing that earlier, done now: https://github.com/simolus3/sqlite-native-libraries/commit/b2933517ceee7fb49d5a9900e6695fe5647da1f3.

I'm in favor of this feature, but I do wonder if we might need to add `SQLITE_CORE` as a define while compiling these extensions? So far, I haven't really used those features of sqlite3, but from [this section](https://www.sqlite.org/loadext.html#statically_linking_a_run_time_loadable_extension) it sounds like that might be helpful for `SQLITE_EXTENSION_INIT1` to become a no-op since we won't ever load it dynamically?

FaFre:
I think we don't have to care about that. The extension itself is not different from the already included ones like json1, rtree or fts5. It's just that the source file is not included by default. The structure of the extension is the same. I think this is a historical part of the documentation, when there were no extensions really added yet.

Also see: https://sqlite-users.sqlite.narkive.com/rep6Tusc/sqlite-core-use-for

FaFre:
I did some more investigation and indeed, it seems to only affect internal extensions and the way the are build (like stated by the creator https://sqlite-users.sqlite.narkive.com/rep6Tusc/sqlite-core-use-for#post4). When `SQLITE_CORE` is NOT set, only pure SQLite is available and internal modules have to get loaded manually. If not, the internal extensions are getting compiled to expose a special init function e.g. `sqlite3Fts5Init`for fts5. SQLite is aware of its internal modules then and uses this specific init to load the extension directly.

Also I found the following code:
```c
/*
** The following functions are used to register the module with SQLite. If
** this module is being built as part of the SQLite core (SQLITE_CORE is
** defined), then sqlite3_open() will call sqlite3Fts5Init() directly.
**
** Or, if this module is being built as a loadable extension, 
** sqlite3Fts5Init() is omitted and the two standard entry points
** sqlite3_fts_init() and sqlite3_fts5_init() defined instead.
*/
#ifndef SQLITE_CORE
#ifdef _WIN32
__declspec(dllexport)
#endif
int sqlite3_fts_init(
  sqlite3 *db,
  char **pzErrMsg,
  const sqlite3_api_routines *pApi
){
  SQLITE_EXTENSION_INIT2(pApi);
  (void)pzErrMsg;  /* Unused parameter */
  return fts5Init(db);
}

#ifdef _WIN32
__declspec(dllexport)
#endif
int sqlite3_fts5_init(
  sqlite3 *db,
  char **pzErrMsg,
  const sqlite3_api_routines *pApi
){
  SQLITE_EXTENSION_INIT2(pApi);
  (void)pzErrMsg;  /* Unused parameter */
  return fts5Init(db);
}
#else
int sqlite3Fts5Init(sqlite3 *db){
  return fts5Init(db);
}
#endif
```
So it seems like `SQLITE_CORE` is already set, since `sqlite3_fts5_init` is not getting exported when looking at the symbols of the compiled library. Also, nowhere in the source `sqlite3_fts5_init` is referenced, so it cannot get automatically loaded anywhere.

However, for spellfix `SQLITE_CORE` is not relevant, because it cannot get internally called anyways (since its not built in extension) and we have to call the `sqlite3_spellfix1_init` ourselfs. 

```c
/*
** Register a statically linked extension that is automatically
** loaded by every new database connection.
*/
int sqlite3_auto_extension(
  void (*xInit)(void)
)
```
So, `sqlite3_auto_extension` is especially for statically linked extensions. Right now all dynamic loading is already disabled, because `SQLITE_OMIT_LOAD_EXTENSION` is already set. We don't have to change anything here.

sqlite3_flutter_libs/linux/CMakeLists.txt
```diff
@@ -13,9 +13,17 @@ FetchContent_Declare(
 )
 FetchContent_MakeAvailable(sqlite3)
 
+FetchContent_Declare(
+  spellfix1
+  URL https://raw.githubusercontent.com/sqlite/sqlite/branch-3.39/ext/misc/spellfix.c
+  DOWNLOAD_NO_EXTRACT TRUE
+)
+FetchContent_MakeAvailable(spellfix1)
```
simolus3, line 21: I think we need the same logic for Windows in this repository, right?

Android should be done in the `sqlite-native-libraries` repository. I include that from maven central here because we're not allowed to ship binaries to pub.dev and since users can't be trusted to get their NDK install right (so we can't ship sqlite3 sources).

sqlite3/lib/src/common/constants.dart
```diff
@@ -507,6 +507,12 @@ class SqlFunctionFlag {
   static const SQLITE_INNOCUOUS = 0x000200000;
 }
 
+enum LoadableExtension { spellfix1 }
+
+const loadableExtensionEntrypoints = {
+  LoadableExtension.spellfix1: 'sqlite3_spellfix_init'
+};
```
simolus3, line 514: Conceptually, I feel like the `sqlite3` package shouldn't define these constants. spellfix isn't part of the generic default sqlite3 build that this package is providing Dart APIs for. Since this is an advanced feature, I think we can make `ensureExtensionLoaded` take the entrypoint symbol as a direct string and avoid the `LoadableExtension` enum.

In the `sqlite3_flutter_libs` package, which is guaranteed to have spellfix1 available, we could define helpers like this:

```dart
enum LoadableExtension {
  spellfix1('sqlite3_spellfix_init');

  final String _entrypoint;
  const LoadableExtension(this._entrypoint);
}

extension LoadKnownExtension on Sqlite3 {
  void loadKnownExtension(LoadableExtension e) => ensureExtensionLoaded(e._entrypoint);
}
```

What do you think?

simolus3 (commented):
Just tracking some todos we need here:

1. Probably some tests. In https://github.com/simolus3/sqlite3.dart/pull/117/files, I have defined a custom dynamic extension which I wanted to run. That didn't end up going anywhere, but you can copy the basic approach to define a custom static extension which we'd then compile and load in a unit test. This test only needs to run in the CI (see also the test configuration changes in that PR). Only running this test on Linux and Windows is fine, dynamically loading sqlite3 on macOS is weird.
2. In the integration tests for `sqlite3_flutter_libs`, we can write a test ensuring that `spellfix1` is available.
3. We need to find a solution for iOS and macOS. If you want to I can take a look at contributing this to the pod we're using at the moment. You can also give it a shot of course, I'm not that experienced with CocoaPods either :)

FaFre:
Do you think it will be possible to also load extensions dynamically on Android and iOS?

I found out I need some internal things of the FTS5 extension, for which I wrote an own extension to expose them. Nothing you want to statically link for sure. So maybe it makes sense to go into the direction of dynamic loading here, so spellfix could also packed into an own package.

What do you think?

simolus3:
I tried to implement dynamic loading in https://github.com/simolus3/sqlite3.dart/pull/117#issuecomment-1226523834.

However, we currently omit the extension loading mechanism from our sqlite3 builds, it is also disabled on iOS/macOS by default. Given the security considerations of dynamic extension loading, I don't think that we should expose the automatic interface loading.

I'm totally fine with an API that you give a function pointer to and it loads that as an extension. This is essentially what is `sqlite3_auto_extension` is, and I'm happy to expose it directly. It doesn't even require static linking, you can put your extension into another shared object, do `DynamicLibrary.open` in Dart, lookup the pointer with `dart:ffi` and then call `sqlite3_auto_extension` on that. Again, from a security perspective, I really don't want _this library_ dynamically loading code. Being given a function pointer that _you_ loaded is fine with me.

> So maybe it makes sense to go into the direction of dynamic loading here, so spellfix could also packed into an own package.

That would be an interesting approach. On iOS and macOS, I managed to configure it so that spellfix1 can be included as an optional dependency (https://github.com/clemensg/sqlite3pod/pull/19). So we could have another Flutter package (say `sqlite3_spellfix1_flutter_libs` :D) only providing that extension on iOS/macOS. On Android, we can pull the extension back out of `sqlite3_native_libraries` and also include it with a native build in that package (or perhaps as a Gradle feature variant in `sqlite3_native_libraries`, but I'm not too familiar with those).

FaFre:
Ohh yeah, I totally forgot that we are also able to call the entry point of other modules :D
I changed the extension loading to make it more dynamical and will do some tests on the weekend.

simolus3:
It looks like support for (statically linking) spellfix1 has landed in the sqlite3 pod.

I have just released `3.39.3` of the native libraries for Android. You can try upgrading that and iOS to `3.39.3`. On iOS, you can also add a dependency to the subpod `sqlite3/spellfix1` now.

FaFre:
I checked your latest commit, do you think its okay to pass the dynamic library to an isolate? I Was not sure about thr behaviour here, thats why I used the resolve.

simolus3:
Thinking about this some more, we should have support for automatically loading extensions, but I don't believe we should ship any of them by default.

There are integration tests for this feature so we know it works, but since it is a pretty advanced use case we might as well expect the user to set up the libraries manually.

simolus3 (approved):

simolus3:
> do you think its okay to pass the dynamic library to an isolate?

Hm, `dart:ffi` types can't be used in isolate messages but I'm not sure if there's a use case for sending this class across isolates in the first place? If you're using this package in a a background isolate, couldn't you just call the method on that isolate.

(well technically you could call that method on any isolate, extensions are global and visible across isolates either way)

PR #117, opened 2022-08-24
Add API to load dynamic extensions

simolus3:


simolus3:
Dynamic extensions aren't supported on macOS by default and we currently disable that mechanism with `sqlite3_flutter_libs` as well. Security wise, is probably for the best. Let's not export that feature as an API then.

PR #118, opened 2022-08-24
Implementation of Backup API

FaFre:
This PR adds support for the SQLite Backup API (https://www.sqlite.org/c3ref/backup_finish.html#sqlite3backupinit) for ffi.

It supports backup/restore from/to memory/disk databases. The implementation follows the official examples: https://www.sqlite.org/backup.html

This can be also useful for testing, since its very easy now to load a prepared database from disk into memory and apply additional tests on it.

### DIFF
--- a/sqlite3/lib/src/ffi/impl/database.dart
+++ b/sqlite3/lib/src/ffi/impl/database.dart
@@ -493,7 +493,7 @@ class DatabaseImpl extends Database {
         final remaining = _bindings.sqlite3_backup_remaining(pBackup);
         final count = _bindings.sqlite3_backup_pagecount(pBackup);
 
-        yield 100 * (count - remaining) / count;
+        yield (count - remaining) / count;
 
         if (returnCode == SqlError.SQLITE_OK ||
             returnCode == SqlError.SQLITE_BUSY ||
@@ -523,6 +523,7 @@ class DatabaseImpl extends Database {
 
   /// Impl only, this should be only called right after opening a new in-memory
   /// database
+  @internal
   void restore(Database fromDatabase) {
     if (!isInMemory()) {
       throw ArgumentError(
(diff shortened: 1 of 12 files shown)


sqlite3/lib/src/ffi/impl/database.dart
```diff
@@ -434,4 +434,112 @@ class DatabaseImpl extends Database {
...
+
+        final remaining = _bindings.sqlite3_backup_remaining(pBackup);
+        final count = _bindings.sqlite3_backup_pagecount(pBackup);
+
+        progressCallback?.call(100 * (count - remaining) / count);
+
+        if (returnCode == SqlError.SQLITE_OK ||
+            returnCode == SqlError.SQLITE_BUSY ||
+            returnCode == SqlError.SQLITE_LOCKED) {
+          _bindings.sqlite3_sleep(250);
```
simolus3, line 502: Do you know whether this needs to be synchronous or if we could resume backups across asynchronous suspensions?

If we can get away with it, I think I'd prefer an API that is essentially `Stream<double> backup(Database toDatabase)` to report progress. Whenever we're entering a locked state, can we just do `await Future.delayed(const Duration(milliseconds: 250))` instead? We can't avoid blocking due to IO, but I'd prefer not blocking Dart threads in `sleep` if possible.

FaFre: I also like the `Stream` much more. Done!
I checked the sqlite source and we can replace the sleep it with an delayed future. Also done!

simolus3:
I think the CI is currently broken because Dart stable and Dart dev disagree on formatting. I tried to fix that in 204a3383b5f1da6c35dceda8c219c012e2152d1b, if that fixes the action on `main` you could merge it into your branch as well.

simolus3 (approved):

simolus3:
Thanks!

PR #130, opened 2022-11-08
Enable fts3 and fts4 support in ios. macis, linux & windows

SNNafi:
Enable __FTS3__ & __FTS4__ support in iOS, macOS, Linux & Windows.

### DIFF
--- a/sqlite3_flutter_libs/ios/sqlite3_flutter_libs.podspec
+++ b/sqlite3_flutter_libs/ios/sqlite3_flutter_libs.podspec
@@ -18,6 +18,7 @@ A new flutter plugin project.
   s.dependency 'Flutter'
 
   s.dependency 'sqlite3', '~> 3.39.4'
+  s.dependency 'sqlite3/fts'
   s.dependency 'sqlite3/fts5'
   s.dependency 'sqlite3/perf-threadsafe'
   s.dependency 'sqlite3/rtree'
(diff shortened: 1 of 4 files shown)


SNNafi:
@simolus3 , First of all, thank you for this great library. I understand your concern. I am ok with that. Actually, It took time for me to figure out that, fts3 and fts4 are disabled in the shipped SQLite.
I think it's better, if you add a line in the __README.md__ about it.

simolus3:
> I think it's better, if you add a line in the __README.md__ about it.

Thanks for understanding, and good point about the readme! I've added a quick note about this in d2e0208d47558ed309b6c225fa7a2f4766168df2.

EArminjon:
> > I think it's better, if you add a line in the **README.md** about it.
> 
> Thanks for understanding, and good point about the readme! I've added a quick note about this in [d2e0208](https://github.com/simolus3/sqlite3.dart/commit/d2e0208d47558ed309b6c225fa7a2f4766168df2). If you need help including fts3 or fts4 in your app, I'm happy to provide some pointers.

How to add ftS4 with this package please :'( ?

simolus3:
@EArminjon You'd have to stop depending on `sqlite3_flutter_libs` and instead add the build scripts manually.

1. On iOS and macOS, add a dependency of `sqlite3/fts` to your podspec
2. On Linux and Windows, adapt your CMakeLists to compile sqlite3 as a shared library (you can take [the plugin's](https://github.com/simolus3/sqlite3.dart/blob/main/sqlite3_flutter_libs/linux/CMakeLists.txt) as inspiration, you'd need to add compile-time options to enable FTS).
3. On Android, you can change your Gradle build logic to compile sqlite3 through the NDK, similar to [how its done here](https://github.com/simolus3/sqlite-native-libraries/blob/master/sqlite3-native-library/build.gradle.kts).

PR #137, opened 2022-12-15
SQLCipher 4.5.2 on Android, iOS and macOS

davidmartos96:
Hello Simon!
This updates the version of SQLCipher to 4.5.2 on Android. On macOS and iOS the podfile.lock already will match to this version, but I added the explicit version nonetheless.
For Windows and Linux it looks like it's being compiled against 4.5.1 sources uploaded by you, so I can't change the url since 4.5.2 sources are not uploaded.

https://www.zetetic.net/blog/2022/08/03/sqlcipher-4.5.2-release/

### DIFF
--- a/sqlcipher_flutter_libs/CHANGELOG.md
+++ b/sqlcipher_flutter_libs/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.3
+
+- Upgrade `SQLCipher` to version `4.5.2`.
+
 ## 0.5.2
 
 - This package now works on Windows and Linux too!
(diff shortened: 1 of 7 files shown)


simolus3:
Thanks! Yeah realistically I think I have to provide the sources since there are no prebuilt binaries and it requires TCL to build, which most users probably don't have. I have updated the provided sources now.

I'll have to test this later today, then I'll merge and release this version.

simolus3 (approved):

davidmartos96:
Thank you!

PR #140, opened 2023-02-09
Setup integration tests on Linux

ky1vstar:
At first I wanted to use [xvfb-action](https://github.com/GabrielBB/xvfb-action), but then I've discovered that it is not maintained anymore. So I switched to `xvfb-run` command which is pre-instlled on `ubuntu-latest` according to this [PR](https://github.com/aioutecism/amVim-for-VSCode/pull/315).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -227,13 +227,17 @@ jobs:
       - name: sqlite3 driver tests
         uses: GabrielBB/xvfb-action@v1
         with:
-          run: "flutter pub get && flutter test integration_test"
+          run: |
+            flutter pub get
+            flutter test integration_test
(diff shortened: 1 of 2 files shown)


ky1vstar:
I've also seen `integration_test_macos` failing in the fork repo, but I don't think it is somehow related to the code changes I propose

simolus3 (approved):
Thanks for your contribution, the `xvfb-run` trick is really neat!

> I've also seen `integration_test_macos` failing in the fork repo, but I don't think it is somehow related to the code changes I propose

I agree, your changes couldn't have caused that regression. I'll take a look at that failure later and merge your PR now, having the tests on Linux is a nice improvement.

PR #150, opened 2023-03-26
Fix platform dependent path

michalsrutek:
I know the example is for Linux (`_openOnLinux`), but even then it might be better to use `join()`, so that it can be easily updated for Windows.

On Windows, the problem would be with slashes vs backslashes in the path.


simolus3 (approved):
Thanks!

PR #158, opened 2023-05-09
Add support for Gradle 8+

davidmartos96:
## Description

Hello, currently the flutter plugins don't build when Gradle 8+ is used. That is because the property `namespace` is mandatory.
I included the namespace property with backwards compatibility.

### DIFF
--- a/sqlcipher_flutter_libs/android/gradle/wrapper/gradle-wrapper.properties
+++ b/sqlcipher_flutter_libs/android/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,6 @@
+#Tue May 09 11:17:21 CEST 2023
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.2-all.zip
--- a/sqlcipher_flutter_libs/android/build.gradle
+++ b/sqlcipher_flutter_libs/android/build.gradle
@@ -4,18 +4,18 @@ version '1.0'
 buildscript {
     repositories {
         google()
-        jcenter()
+        mavenCentral()
     }
 
     dependencies {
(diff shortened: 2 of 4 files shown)


sqlcipher_flutter_libs/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Tue May 09 11:17:21 CEST 2023
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-bin.zip
```
simolus3, line 4: Can we upgrade these to the latest 8.1.1 or is there something else causing an incompatibility?

simolus3 (commented):
Thanks!

sqlcipher_flutter_libs/android/gradle/wrapper/gradle-wrapper.properties
davidmartos96: I'm not too experienced with how gradle versions work between the main app and other modules/plugins.
Would using AGP 8 in the plugin break some app using an old version of AGP? Or is Gradle backwards compatible?

I can update to 8.x, it build ok

simolus3 (approved):
I was worried that adding a dependency on the newer android tools plugin would cause problems for users running builds with Gradle 7.x, but that doesn't seem to be the case. Thanks for the contribution :+1:

davidmartos96:
> I was worried that adding a dependency on the newer android tools plugin would cause problems for users running builds with Gradle 7.x, but that doesn't seem to be the case. Thanks for the contribution +1

Yes, I was in the same boat, and since the official Flutter plugins were using 7.x in the AGP even after adding support to Gradle 8, I did the same. I'm not sure how the version of the AGP in each module affect the version chosen in the apps

davidmartos96:
@simolus3 Apparently using Gradle 8 in the plugin build.gradle may cause issues to some users.
Here is an example from another package where the AGP version was bumped to 8.x
https://github.com/MisterJimson/flutter_keyboard_visibility/issues/132

The test I did with 7.x in the client side worked fine. Maybe the error is when using even lower versions.

Issue #171, opened 2023-07-27
SQLCipher compile flags consistency


davidmartos96:
Hello!
I was looking into using the linux library for an existing Android/iOS app that uses `sqlcipher_flutter_libs`.
I've found that the flags used on Linux and Windows have some differences between iOS and Android that can break an existing app implementation.

Here are the flags used on the dependencies maintained by the SQLCipher team (Android, iOS and macOS) 
https://github.com/sqlcipher/sqlcipher/blob/master/SQLCipher.podspec.json
```json
"compiler_flags": [
        "-DNDEBUG",
        "-DSQLITE_HAS_CODEC",
        "-DSQLITE_TEMP_STORE=2",
        "-DSQLITE_SOUNDEX",
        "-DSQLITE_THREADSAFE",
        "-DSQLITE_ENABLE_RTREE",
        "-DSQLITE_ENABLE_STAT3",
        "-DSQLITE_ENABLE_STAT4",
        "-DSQLITE_ENABLE_COLUMN_METADATA",
        "-DSQLITE_ENABLE_MEMORY_MANAGEMENT",
        "-DSQLITE_ENABLE_LOAD_EXTENSION",
        "-DSQLITE_ENABLE_FTS4",
        "-DSQLITE_ENABLE_FTS4_UNICODE61",
        "-DSQLITE_ENABLE_FTS3_PARENTHESIS",
        "-DSQLITE_ENABLE_UNLOCK_NOTIFY",
        "-DSQLITE_ENABLE_JSON1",
        "-DSQLITE_ENABLE_FTS5",
        "-DSQLCIPHER_CRYPTO_CC",
        "-DHAVE_USLEEP=1",
        "-DSQLITE_MAX_VARIABLE_NUMBER=99999"
],
```

## Differences found

1. THREADSAFE is 2 instead of 1 like on `sqlite3_flutter_libs` and like on the team maintained flags. Related https://github.com/simolus3/drift/issues/2032#issuecomment-1247186128
2. `sqlite3_flutter_libs` and `sqlcipher_flutter_libs` (linux and windows) use the flag `DQS=0` which disables the usage of double quotes as string literal. The SQLCipher maintained libs don't set that flag, so existing apps that have used the Android and iOS dependency may use double quotes in many different queries. Using the Linux or Windows dependency now breaks those queries.

1 is an easy fix, like it was done for `sqlite3_flutter_libs'  v0.5.10

For 2 I can see two solutions.

[...]

PR #172, opened 2023-07-27
Add option to enable/disable double quotes as literals

davidmartos96:
This could be a solution for https://github.com/simolus3/sqlite3.dart/issues/171
It exposes the `sqlite_db_config` function that we can use to configure the double quotes for a particular database connection.

I added a test for the setting but docs are missing.
Let me know if the implementation and code organization seems ok to you and I'll complete it with documentation if you want to go forward with this solution.

### DIFF
--- a/sqlite3/lib/src/database.dart
+++ b/sqlite3/lib/src/database.dart
@@ -8,7 +8,12 @@ import 'constants.dart';
 /// An opened sqlite3 database.
 abstract class CommonDatabase {
   /// Configuration for the database connection.
-  DatabaseOptions get options;
+  ///
+  /// __Note__: On the web, the [DatabaseConfig] class only works when using a
+  /// version of `sqlite3.wasm` shipped with version 2.1.0 of the `sqlite3`
+  /// Dart package. In previous WASM builds, all setters on the config will
+  /// throw an exception.
+  DatabaseConfig get config;
 
   /// The application defined version of this database.
   abstract int userVersion;
@@ -212,7 +217,7 @@ final class SqliteUpdate {
 ///
 /// More information: https://www.sqlite.org/c3ref/db_config.html
 /// Available options are documented in https://www.sqlite.org/c3ref/c_dbconfig_defensive.html
-abstract base class DatabaseOptions {
+abstract base class DatabaseConfig {
   /// Update configuration that accepts an int value.
   /// Would throw when the internal C call returns a non-zero value.
   void setIntConfig(int key, int configValue);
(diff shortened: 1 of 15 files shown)


sqlite3/lib/src/database.dart
```diff
@@ -157,6 +157,8 @@ abstract class CommonDatabase {
     bool directOnly = true,
   });
 
+  void configDoubleQuotedStringLiterals({required bool enable});
```
simolus3, line 160: Instead of exposing this on the database directly, I think it could make sense to have a class exposing the available options.

E.g. the database class could have a `DatabaseOptions get options;` getter, where `DatabaseOptions` would look like this:

```dart
abstract base class DatabaseOptions {
  void setIntConfig(int key, int configValue); // Would throw when the internal C call returns a non-zero value

  set doubleQuotedStringLiterals(bool value) {
    setIntConfig(SQLITE_DBCONFIG_DQS_DML, value ? 1 : 0);
    setIntConfig(SQLITE_DBCONFIG_DQS_DQL, value ? 1 : 0);
  }
}
```

This would allow us to export additional options in the future without cluttering the `Database` interface.

davidmartos96: Yes, good idea, I changed it.
Should we use `late final` in the subclass for the options getter?

simolus3 (approved):
I had to write a wrapper for WASM, I couldn't figure out what the calling convention for variadic functions is there. Apparently we're supposed to allocate the variadic structure on memory somewhere, but it looks like the exact layout may not be properly specified.

I've renamed `options` to `config` as well to be consistent with the sqlite3 API - sorry for suggesting `options` originally.

davidmartos96:
Great, thanks Simon!
One question from the changes. Wasn't setting 3 or 0 based on the boolean correct?
I based it on the table from https://www.sqlite.org/compile.html#dqs

simolus3:
The 2-bit value from the compile-time option is used to control the value for both the DDL and the DQL. In the config interface, we have two different keys which take boolean values.
It works with `3` as well since sqlite3 takes any non-zero value as true, but I think using 0/1 captures the type better.

davidmartos96:
> The 2-bit value from the compile-time option is used to control the value for both the DDL and the DQL. In the config interface, we have two different keys which take boolean values. It works with `3` as well since sqlite3 takes any non-zero value as true, but I think using 0/1 captures the type better.

Oh yes, you are right. Here we are setting them individually, so that's not needed.

PR #185, opened 2023-09-28
Use native assets experiment to compile sqlite3

simolus3:
Use the `--native-assets` experiment to compile sqlite3 without using platform-specific build scripts (like `sqlite3_flutter_libs`).

We still need to support downloading sqlite3 from an URL as a build mode. It would also be nice to be able to pass additional compile flags or remove to some of the `OMIT`s we're applying.

Another todo: We're now able to avoid some of the copies when binding blobs and strings by directly passing the list.


simolus3:
Had to abandon this approach, but https://github.com/simolus3/sqlite3.dart/pull/285 works and is almost ready to be released.

PR #189, opened 2023-10-19
Fix web support docs, add `example/multiplatform`

denis-ryzhkov:
* Fixed "WASM (web support)" section of `README`:
  * Replaced `SqliteEnvironment` (which is gone from the codebase) with `registerVirtualFileSystem`.
  * Fixed [IndexedDbFileSystem.open](https://pub.dev/documentation/sqlite3/latest/sqlite3.wasm/IndexedDbFileSystem/open.html) call which requires named argument `dbName` instead of positional one.
* Added `example/multiplatform` which uses common interface to `sqlite3` on web and native platforms.
  * To run this example, merge its files into a Flutter app.

### DIFF
--- a/sqlite3/example/multiplatform/db/sqlite3/web.dart
+++ b/sqlite3/example/multiplatform/db/sqlite3/web.dart
@@ -3,6 +3,8 @@ import 'package:sqlite3/wasm.dart' show IndexedDbFileSystem, WasmSqlite3;
 
 Future<CommonDatabase> openSqliteDb() async {
   const name = 'my_app';
+  // Please download `sqlite3.wasm` from https://github.com/simolus3/sqlite3.dart/releases
+  // into the `web/` dir of your Flutter app. See `README.md` for details.
   final sqlite = await WasmSqlite3.loadFromUrl(Uri.parse('sqlite3.wasm'));
   final fileSystem = await IndexedDbFileSystem.open(dbName: name);
   sqlite.registerVirtualFileSystem(fileSystem, makeDefault: true);
--- a/sqlite3/example/multiplatform/db/sqlite3/web.dart
+++ b/sqlite3/example/multiplatform/db/sqlite3/web.dart
@@ -1,0 +1,10 @@
+import 'package:sqlite3/common.dart' show CommonDatabase;
+import 'package:sqlite3/wasm.dart' show IndexedDbFileSystem, WasmSqlite3;
+
+Future<CommonDatabase> openSqliteDb() async {
+  const name = 'my_app';
+  final sqlite = await WasmSqlite3.loadFromUrl(Uri.parse('sqlite3.wasm'));
+  final fileSystem = await IndexedDbFileSystem.open(dbName: name);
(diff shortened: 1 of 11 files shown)


sqlite3/example/multiplatform/db/sqlite3/web.dart
```diff
@@ -0,0 +1,10 @@
+import 'package:sqlite3/common.dart' show CommonDatabase;
+import 'package:sqlite3/wasm.dart' show IndexedDbFileSystem, WasmSqlite3;
+
+Future<CommonDatabase> openSqliteDb() async {
+  const name = 'my_app';
+  final sqlite = await WasmSqlite3.loadFromUrl(Uri.parse('sqlite3.wasm'));
```
simolus3, line 6: I think it may be helpful to mention that this requires an additional file to be added to the project, for instance with a comment pointing back to the readme.

simolus3 (approved):
Thanks for the fixes to the readme and the example! This looks good to me, but let's add a comment explaining that `sqlite3.wasm` needs to be downloaded (so that it's hopefully more obvious for someone copying these files).

sqlite3/example/multiplatform/db/sqlite3/web.dart
denis-ryzhkov: Thanks, done: https://github.com/simolus3/sqlite3.dart/pull/189/commits/1d436749fd7915c9cf2254bb2a1309ed4c513890

denis-ryzhkov:
* CI needs `path_provider` to analyze `example/multiplatform`
* Error: https://github.com/simolus3/sqlite3.dart/actions/runs/6584478470/job/17889282130?pr=189#step:8:10
* Fix: https://github.com/simolus3/sqlite3.dart/pull/189/commits/658183c687bc93d351f26ad9a35f88f833f6ce2a

sqlite3/example/multiplatform/pubspec.yaml
```diff
@@ -0,0 +1,16 @@
...
+description: Uses common interface to `sqlite3` on web and native platforms.
+version: 0.1.0
+publish_to: 'none'
+
+environment:
+  sdk: ">=2.12.0 <4.0.0"
+
+dependencies:
+  path: ^1.8.3
+  path_provider: ^2.1.1
```
simolus3, line 11: Installing the Flutter SDK in GitHub actions takes much longer than a plain Dart SDK, but it seems like we can't resolve these dependencies otherwise.

Can we comment this out as well and change the analysis step to only run on `lib/` and `test/` instead? The command in line 116 of the action could be replaced with `dart analyze --fatal-infos lib/ test/`

denis-ryzhkov: Right, we need to skip CI `analyze` of example which requires `path_provider` which requires Flutter.

Done: https://github.com/simolus3/sqlite3.dart/pull/189/commits/58338bc591344d58445519e1d4608f55c3acc064

simolus3:
Thanks!

PR #201, opened 2023-12-18
Expose sqlite3_get_autocommit

rkistner:
Expose [sqlite3_get_autocommit](https://www.sqlite.org/c3ref/get_autocommit.html) to check whether or not the connection is currently within a transaction.

For the use case, see https://github.com/powersync-ja/sqlite_async.dart/issues/21.

Note: I added an implementation for both ffi and wasm, but I haven't tested the wasm version (couldn't quickly see how to run the tests).


rkistner:
> The easiest way is to download sqlite3.wasm from [here](https://github.com/simolus3/sqlite3.dart/releases) and place it to example/web

Great, that worked. Tests are passing for wasm as well.

simolus3 (approved):
Thanks!

simolus3:
Released in version 2.3.0.

PR #203, opened 2024-01-01
Update README.md

niku:
On the latest version v2.3.0, I couldn't find API `FileSystem.inMemory`.
It seems it was replaced with `InMemoryFileSystem`.


simolus3 (approved):
Thanks!

PR #210, opened 2024-02-14
Bump js dependency to ^0.7.0

rkistner:
This resolves the "Support up-to-date dependencies" scoring issue on pub.dev for this package and others that depend on it. Tests still pass on my machine for native and wasm.

Update: I see the issues with js `^0.7.0` and integration_test now, `>=0.6.4 <0.8.0` should work to still support that.


simolus3 (approved):
The web tests still pass with the upgraded version, so this looks good to me - thanks!

PR #232, opened 2024-05-19
Better wasm support

simolus3:
- As `dart2wasm` can't convert JS typed data instances into Dart instances without copying, we have to avoid `dart:typed_data` pretty much everywhere in code we intend to compile to WASM. This affects the VFS layer which is a breaking change.
- todo there will probably be more changes required.

### DIFF
--- a/sqlite3/lib/common.dart
+++ b/sqlite3/lib/common.dart
@@ -10,4 +10,3 @@ export 'src/result_set.dart';
 export 'src/sqlite3.dart';
 export 'src/statement.dart'
     show CommonPreparedStatement, StatementParameters, CustomStatementParameter;
-export 'src/vfs.dart';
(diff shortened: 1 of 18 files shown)


hnvn:
I've got this error when trying to run `flutter build web --wasm`:
```
../../../.pub-cache/hosted/pub.dev/sqlite3-2.4.6/lib/src/ffi/memory.dart:2:1: Error: 'dart:ffi' can't be imported when compiling to Wasm.
import 'dart:ffi';
```

hnvn:
Thank you @simolus3 , I get it work now, it's because of using old implementation that import `sqlite3_flutter_libs` manually. I followed new setup docs to add dependencies for `drift` and `drift_flutter`.

PR #236, opened 2024-06-02
Enable extension loading

mugikhan:
This PR enables manual extensions loading via manual C-API call (https://www.sqlite.org/c3ref/enable_load_extension.html) for the built binaries on both Windows and Linux platforms.

Following on from the discussion on PR #178, this makes the necessary changes for consistency on all platform builds excluding wasm builds (not supported at the moment).


mugikhan:
@simolus3 Here is a PR I created for enabling extension loading on Windows and Linux. Please review it and let me know if I missed something.

simolus3:
Merged and published, thanks for the contribution!

PR #256, opened 2024-10-12
Adding uuid.c example

rodydavis:
Added example for how to compile a sqlite c extension for use in Flutter via FFI plugin template.

Web not supported since it would need to be compiled into WASM build.


simolus3:
Thanks! I've also linked this example from the `SqliteExtension` class.

rodydavis:
Thanks! It was challenging to figure out Cmake but thought it could help others 👍🏼

PR #258, opened 2024-10-18
Add a DynamicBuffer for faster VFS implementations

rkistner:
Currently, the InMemoryFileSystem uses a Uint8List buffer per file. When resizing the buffer, it creates a new one and copies the content over.

While copying the bytes is fairly fast for small buffers, doing that for every written page for a 10MB+ file size adds significant overhead. Basically, creating a database with size n takes `O(n^2)` time, which becomes very noticeable with large databases. This affects both IndexedDbFileSystem and InMemoryFileSystem, since IndexedDbFileSystem uses InMemoryFileSystem internally. In some of my tests, creating a 30MB database takes around 120s with either VFS.

This introduces a DynamicBuffer class that handles resizing the buffer. Instead of resizing on every write, it doubles the size every time new capacity is needed. This changes it to `O(n)` to create a database of size n. In my tests, the same database takes 5s to create with InMemoryFileSystem, and 10s with IndexedDbFileSystem.

Another approach would be to use the JS [ArrayBuffer.resize](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/resize) together with `maxByteLength`, since ArrayBuffers are used internally for ByteBuffer with dart2js anyway. However, it may not be supported by all browsers, and would need a different approach for dart2wasm.

In my testing this appears to work well with both IndexedDbFileSystem and InMemoryFileSystem. It would still be good to review the IndexedDbFileSystem implementation specifically - I'm not sure whether the change could introduce some edge case, especially around `previousContent`.


simolus3 (commented):
Thanks! I agree with the better management for handling growing data.

If possible, I prefer using an `Uint8Buffer` from `package:typed_data` for this though. That class is supposed to be doing the same thing, it's a core Dart package and one less thing to maintain here :)

rkistner:
Great, I wasn't aware of that package - switched over now.

simolus3:
Thanks! I've merged this and will upload a this weekend.

Issue #265, opened 2024-11-21
[feat] Add commit hook support


jackd:
Love the package/ecosystem. One piece of functionality I'd love is access to [commit hooks](https://www.sqlite.org/c3ref/commit_hook.html).

```c
void *sqlite3_commit_hook(sqlite3*, int(*)(void*), void*);
void *sqlite3_rollback_hook(sqlite3*, void(*)(void *), void*);
```

If there's interest in supporting it, I could have a stab at putting together a PR based on `sqlite3_update_hook`

simolus3:
> If there's interest in supporting it, I could have a stab at putting together a PR based on `sqlite3_update_hook`

A PR would definitely be welcome! Given that commit and rollback hooks also aren't allowed to modify the database, exposing them as streams similar to how update hook work sounds good to me.

Working on the FFI bindings is a bit complicated and we'd also have to support this for WebAssembly, but if you want to get started on this I'm happy to give pointers or help with code if you get stuck somewhere.

jackd:
Terribly sorry, work got hectic for a bit, then Christmas... and now I'm heading away for a few weeks. Just popping by to say this'll be near the top of the list of things I want to get onto when I'm back :)

jackd:
Added in #268

PR #268, opened 2024-11-25
Added commit / rollback hooks

jackd:
As requests in #265, I couldn't get any web tests working so I wouldn't at all be surprised if they're broken... but might be best to seek feedback on the overall design before getting too finicky there.


sqlite3/assets/wasm/bridge.h
```diff
@@ -50,10 +50,12 @@ import_dart("function_xInverse") extern void dartXInverse(
 import_dart("function_xFinal") extern void dartXFinal(sqlite3_context *ctx);
 import_dart("function_xValue") extern void dartXValue(sqlite3_context *ctx);
 import_dart("function_forget") extern void dartForgetAboutFunction(void *ptr);
-import_dart("function_hook") extern void dartUpdateHook(void *id, int kind,
-                                                        const char *db,
-                                                        const char *table,
-                                                        sqlite3_int64 rowid);
+import_dart("function_update_hook") extern void dartUpdateHook(void *id, int kind,
```
simolus3, line 53: Please keep the existing name unchanged, newer versions of `package:sqlite3` need to be compatible with older WebAssembly bundles too.

sqlite3/lib/src/implementation/database.dart
```diff
@@ -228,6 +231,8 @@ base class DatabaseImplementation implements CommonDatabase {
       listener.close();
     }
```
simolus3, line 232: I think we should also close the rollback listeners here.

sqlite3_web/lib/src/database.dart
```diff
@@ -50,6 +50,12 @@ abstract class Database {
   /// stream is active.
   Stream<SqliteUpdate> get updates;
 
+  /// A relayed stream of events triggered by rollbacks from the remote worker.
+  ///
+  /// Updates are only sent across worker channels while a subscription to this
+  /// stream is active.
+  Stream<void> get rollbacks;
```
simolus3, line 57: It's not clear how we would best translate this feature to `sqlite3_web`. We can't provide commit hooks that would be able to revert a transaction because that would require an async roundtrip between the worker and the main page.

We could provide a "read-only" `Stream<void> get commits` stream that installs a `() => true` filter reporting events if necessary. But I'm also fine with not exposing this in `sqlite3_web` until it is needed (`sqlite3_web` also provides the option of compiling custom workers where users have full control over the database, so they could install commit / rollback filters on the database if they need to).

jackd:
Terribly sorry, work got hectic for a bit, then Christmas... and now I'm heading away for a few weeks. Just popping by to say this'll be near the top of the list of things I want to get onto when I'm back :)

jackd:
@simolus3 I've reverted the breaking name change and added a `commits` stream that should be compatible with a custom `commitFilter`. I believe the issue I'm having running web tests is down to me being on WSL. I'm not going to pretend to understand what's going on with the wasm code, just tried to make it look similar to updates... but that's obviously not working.

simolus3:
Thanks for following up on this! The web tests will require a new version of the WebAssembly module. We strip functions that we aren't calling in Dart out of the module to reduce its size. So every time that we add new functions possibly called in Dart, we need to preserve backwards compatibility with older modules that don't have those functions.

The setup to compile the modules is quite complex and I have no idea on how it might be done on Windows. I'll compile a module that has those functions, that was we can get the web part ready.

simolus3 (approved):
LGTM! I've moved some bits around to re-use code between the different types of streams.

Thank you for your contribution!

PR #269, opened 2024-11-25
Fixed bug when setting update_hook to null twice

jackd:
I don't believe this was causing any issues, but it looked smelly.


simolus3 (approved):
Thanks!

PR #276, opened 2025-01-16
`sqlite3_web`: Fix handling worker errors

simolus3:
It's not that constructing the `Worker` throws, an error event is emitted that is currently not handled (meaning that the initialization future never completes). This refactors that to throw an error that is then caught and handled.


dickermoshe:
@simolus3 Seeing much more activity now.
Is working on drift now part of your day job?

PR #290, opened 2025-03-06
Try loading winsqlite3.dll on Windows

benthillerkus:
Windows ships with a sqlite3 dll under the name winsqlite3 since a couple of years. Similar to macOS and Linux this can be used, if no override has been provided

(I accidentally closed #289 by renaming the branch)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -195,7 +195,7 @@ jobs:
     - name: Test sqlite3 package
       run: |
         dart pub get
-        dart test -P ci
+        dart test --debug -P ci
       working-directory: sqlite3/
 
     - name: Test with SQLITE_OMIT_AUTOINIT
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -212,7 +212,7 @@ jobs:
       if: matrix.sqlite != 'system' || runner.os != 'Windows'
       run: |
         dart pub get
-        dart test --debug
+        dart test
       working-directory: sqlite3_test/
 
     - name: Web tests
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -209,6 +209,7 @@ jobs:
       working-directory: sqlite3/
 
     - name: Test sqlite3_test package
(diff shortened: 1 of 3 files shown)


simolus3 (commented):
Thank you for this! I agree that using `winsqlite3.dll` as a fallback is good, but:

1. There should be a comment explaining that this is the name of the shared library on windows, perhaps also mentioning [this link](https://renenyffenegger.ch/notes/Windows/dirs/Windows/System32/winsqlite3_dll/PowerShell/index) which seems to be the best source I could find on this.
2. It would be convenient to test this as part of our CI. Can you see if [copying this](https://github.com/simolus3/sqlite3.dart/blob/a42bf40fa90ca68aa62a8661b69a96e0d57ad972/.github/workflows/main.yml#L134-L178) (while removing the other operating systems as well as the steps to download and install the prebuilt sqlite3 binaries) still passes?

benthillerkus:
> 1. There should be a comment explaining that this is the name of the shared library on windows, perhaps also mentioning [this link](https://renenyffenegger.ch/notes/Windows/dirs/Windows/System32/winsqlite3_dll/PowerShell/index) which seems to be the best source I could find on this.

Added a comment. Winsqlite seems to be kinda undocumented by MS, but it should be fine; they cannot just remove it without breaking compatibility.

I kept the code that tries to open `sqlite3.dll` first, because maybe someone was relying on this (and also I have no clue how `sqlite3_flutter_libs` works, and maybe that package needs that).

> 2. It would be convenient to test this as part of our CI.

I added a new dimension to the test matrix for system provided dlls vs the compiled ones.
There is some weirdness with winsqlite as you can maybe see in the commit messages.

The tests crash when trying to call the function to register a virtual file system. Atleast in CI, I cannot reproduce this on my Windows 11 machine locally.

For some reason either using --debug or enabling randomized test ordered made the regular test suite pass in CI. This doesn't work for the package with test helpers, so I ended up disabling that for winsqlite testing. I'm kind of out of my depth here.

simolus3 (approved):
I'm fine with skipping tests on `sqlite3_web` if they're failing with the system's build on Windows. Thanks for your contribution 👍

PR #316, opened 2025-08-06
Add `updatesSync` stream

simolus3:
This adds a synchronous variant of the `CommonDatabase.update` streams to avoid internal buffers for large transactions.


simolus3:
Rebased and released.

PR #326, opened 2025-10-31
Link math for Android

mattbajorek:
## Reasoning
- Android will not compile with default settings without math linked


simolus3:
Thanks! I'm happy to merge and release this. But as you can tell from the fact that this wasn't done before, `sqlite3_native_assets` is not that well tested across platforms. I'm currently working on integrating native assets into the main `sqlite3` package (via the 3.x beta releases), and those already contain this fix (and have actual integration tests on emulators). `sqlite3_native_assets` will be discontinued in favor of the main package once Dart 3.10 is stable.

PR #337, opened 2025-12-06
web: allow executing multiple statements

stevenctl:
Adds a method that allows executing multiple statements from a single sql string. 
See https://github.com/powersync-ja/sqlite_async.dart/pull/123 and https://github.com/powersync-ja/sqlite_async.dart/issues/122 for context.

Disclaimer: I have not taken the time to set up the test dependencies on my my machine yet.

### DIFF
--- a/sqlite3_web/lib/src/client.dart
+++ b/sqlite3_web/lib/src/client.dart
@@ -150,6 +150,29 @@ final class RemoteDatabase implements Database {
   }
 
   @override
+  Future<void> executeMultiple(
+    String sql, {
+    bool checkInTransaction = false,
+    LockToken? token,
+    Future<void>? abortTrigger,
+  }) async {
+    await connection.sendRequest(
+      RunQuery(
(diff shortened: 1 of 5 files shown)


simolus3 (commented):
Thanks, I agree this functionality is a useful addition 👍 

Implementation-wise, I think we can probably do this within the existing protocol. When running without parameters, `CommonDatabase.execute` will use [`sqlite3_exec`](https://github.com/simolus3/sqlite3.dart/blob/a6ebb6eb471a49bb84e1077092e648f4a631f143/sqlite3/lib/src/implementation/database.dart#L276-L290) which internally [supports multiple statements](https://github.com/sqlite/sqlite/blob/989e8933920fb0c66361b97519cb4fac7fdb4d6f/src/legacy.c#L113-L114).

Given that the worker already uses `execute`, I'd expect that calling `execute` on the client without parameters should accept multiple statements. Is that not what you were seeing?

PR #375, opened 2026-05-21
SQLCipher build hook

davidmartos96:
Questions:

Should we reuse the download_sqlite and build_sqlite jobs?

Regarding build flags, what should we use as default? 
1. The same ones as the current build hook + minimal extra required flags 
2. The same set of build flags that the SQLCipher community build includes https://github.com/sqlcipher/sqlcipher-android/blob/7fab57af75039e5004b087086142b11a9d2a2380/sqlcipher/src/main/jni/sqlcipher/Android.mk#L9


davidmartos96:
@simolus3 , I left a few general questions about how to aproach the PR

davidmartos96:
@simolus3 This is ok to review now. But it appears to be stuck in the integration tests, but I'm not sure why, as I haven't touched those.

sqlite3/hook/build.dart
```diff
@@ -46,7 +50,7 @@ void main(List<String> args) async {
         // For the full discussion, see https://github.com/dart-lang/native/issues/2724
 
         String? linkerScript;
-        if (input.config.code.targetOS == OS.linux) {
+        if (targetOS == OS.linux && libraryType != LibraryType.sqlcipher) {
```
simolus3: Does the linker script or `-Bsymbolic` cause issues with SQLCipher? Why aren't we using it?

sqlite3/lib/src/hook/asset_hashes.dart
```diff
@@ -6,43 +6,19 @@
 
 // dart format off
 // ignore: unnecessary_nullable_for_final_variable_declarations
-const String? releaseTag = 'sqlite3-3.3.1';
+const String? releaseTag = 'sqlite3-sqlcipher-3.3.2';
 
 const Map<String, String> assetNameToSha256Hash = {
```
simolus3, line 11: We would have to include the old hashes again after we have a good CI run.

sqlite3/lib/src/hook/openssl.dart
```diff
@@ -0,0 +1,411 @@
...
+import 'package:code_assets/code_assets.dart';
+import 'package:hooks/hooks.dart';
+import 'package:path/path.dart';
+import 'package:sqlite3/src/hook/android_ndk.dart';
+
+// Based from https://github.com/LucazzP/openssl_dart
+
+// Build instructions: https://github.com/openssl/openssl/blob/openssl-3.6.2/INSTALL.md#building-openssl
+
+const configArgs = [
```
simolus3: I don't think this should be part of the `sqlite3/` package. Can we change the setup to download and build OpenSSL as a static library with the scripts in the top-level `tool/` directory? Then the actual build hook building SQLCipher could expect a static library to exist already instead of compiling it.

Building it can also be a shell script fwiw, we just need some way to build it in CI and not necessarily something that users can run with Dart on their machine.

sqlite3/lib/src/hook/openssl.dart
```diff
@@ -0,0 +1,411 @@
...
+    (OS.windows, Architecture.ia32) => 'VC-WIN32',
+    (OS.windows, Architecture.x64) => 'VC-WIN64A',
+
+    _ => throw UnsupportedError(
+      'Unsupported target combination: ${os.name}-${architecture.name}',
+    ),
+  };
+}
+
+Future<Map<String, String>> resolveWindowsBuildEnvironment(
```
simolus3: Why can't use the compiler we get from `native_toolchain_c`?

tool/build_sqlite.dart
```diff
@@ -12,6 +12,10 @@ import '../sqlite3/hook/build.dart' as hook;
 
 final _limitConcurrency = Pool(Platform.numberOfProcessors);
 
+const _kSQLiteMode = 'sqlite3';
+const _kSQLite3MCMode = 'sqlite3mc';
+const _kSQLCipherMode = 'sqlcipher';
```
simolus3: nit: Avoid the `k` prefix

simolus3 (commented):
> Should we reuse the download_sqlite and build_sqlite jobs?

Yes, `download_sqlite` should also download OpenSSL and SQLCipher sources. `build_sqlite` should also build SQLCipher.

> Regarding build flags, what should we use as default?

Using options from our current build hook + the extra init/shutdown flags sounds correct to me. There's no reason to enable fts4 or extension loading through SQL functions.

sqlite3/lib/src/hook/openssl.dart
davidmartos96: The openssl part was adapted from the openssl dart package, which uses a build hook, I didn't look into the details, it probably can be improved. I didn't like the part that finds the Android NDK either.

Should we do the build openssl in the tool directory in Dart?

sqlite3/lib/src/hook/openssl.dart
davidmartos96: What's the reasoning for not compiling it from the build hook? Isn't it more flexible as it is right now? Having the target OS and architecture is a plus.
The current code already builds a static library for Android and Linux

simolus3: IMO, it feels wrong for build hooks to download external files. Obviously the default hook in the SQLite package is pushing that a bit, but it downloads from a single source and pins hashes of downloaded assets (so there's a guarantee that I can't mess with the assets after publishing to pub.dev).

Having more logic in `tool/` should help (we should be able to use `testBuildHook` with a custom function from `native_toolchain_c` to resolve the default compiler for the target OS + arch too).

sqlite3/lib/src/hook/openssl.dart
simolus3: > Should we do the build openssl in the tool directory in Dart?

I think that could be reasonable, yeah.

sqlite3/lib/src/hook/openssl.dart
davidmartos96: > testBuildHook with a custom function from native_toolchain_c to resolve the default compiler for the target OS + arch

I'm not sure what you mean here. 

Just to clarify, the openssl download would only occur if someone uses the source mode option. What you are suggesting is that when using the source option, the end user should ensure that openssl is compiled for each arch in some place in their file system? If that's the case, I'm not sure how a muliplatform openssl-install dir should be implemented, as the sqlcipher build would need to know where the openssl binaries are located for each arch.

Could you provide an example of a build hook configuration for sqlcipher should look like for the end user? Both compile and prebuilt configurations 

Thanks!

simolus3: Sorry, what I meant to suggest is that we compile SQLCipher for each supported platform as part of the CI in this package (which is the same thing we already do for plain SQLite and SQLite3 Multiple Ciphers). For released packages, the only thing the build hook would do is to download a prebuilt SQLCipher library attached to a release on this package.

This would give us more flexibility for the build, since it can be a collection of shell scripts or other things that only need to work in GitHub actions. I can get started with an example on how I'd structure that if it helps.

davidmartos96: That's how I did it actually 😅
The `sqlcipher` mode in the hook descriptor downloads the prebuilt from GitHub

simolus3: Oh right, I need to take a more in-depth look. But then I assume it's mostly a question of moving this out of the package and into `tool/`. Also things like installing perl and jom should just happen through GitHub actions steps instead.

davidmartos96: Yes, we can move it. Maybe now you understand better my questions from earlier in the conversation.

Prebuilt mode is pretty straightforward, but not sure how should the compile from source option behave for the end user, because of the external dependencies. I like that the hook already compiles the openssl dependency and links it statically with the options to keep it as slim as possible.

> Just to clarify, the openssl download would only occur if someone uses the source mode option. What you are suggesting is that when using the source option, the end user should ensure that openssl is compiled for each arch in some place in their file system? If that's the case, I'm not sure how a muliplatform openssl-install dir should be implemented, as the sqlcipher build would need to know where the openssl binaries are located for each arch.

> Could you provide an example of a build hook configuration for sqlcipher should look like for the end user? Both compile and prebuilt configurations

simolus3: Given how specialized the build is, I don't think we need a compile from source option for end users (maybe that's where the confusion comes from). The existing compile from source option relies on users providing a amalgamation that can be compiled without additional dependencies, so SQLCipher is not suitable for that.

davidmartos96: @simolus3 Yes, I was trying to make it the same way as SQLite. But actually, SQLCipher also supports an amalgamation option by doing `make sqlite3.c`, which I'm not currently using in the PR, but we could. I took the route of downloaidng the already amalgamated source code https://github.com/chehrlic/sqlcipher-amalgamation.   I think the only difference is the external dependency (OpenSSL).

If we want to take only the prebuilt mode route, how should we define the hook input? I was currently using the same "`source': 'source'` hook define as an entrypoint.

simolus3: I've implemented a few of the things I had in mind on the `sqlcipher` branch, could you use that one as a target?

Basically, my idea is:

1. The download step should also build the amalgamation for SQLCipher (which requires tcl, but that's installed on GH actions by default).
2. I don't think it makes sense to cache OpenSSL downloads since we can download from GH releases which is fairly quick. We do want to build OpenSSL libraries in a separate job though (for more fine-grained cache invalidation), and upload static OpenSSL libraries as an artifact.
3. The SQLite build step can use those artifacts. We can introduce an internal user-define read by the hook and set in `tool/build_sqlite.dart` to make it link the relevant frameworks/OpenSSL builds.

I've completed step 1. Currently I'm only building SQLCipher on macoS (and on Linux but with a dynamic `-lcrypto` from the OS).  I've also set up step 2, but for Linux only.

I won't have much more time for this until next week, I think a PR for any of the following would help:

1. Porting the invocation to build OpenSSL on Windows and Android (but install dependencies via GitHub actions, not in a Dart script - similar to how we install cross-compiling GCCs via GH actions). You can probably copy this from what you already have.
2. Figuring out why Linux cross-compilation for arm and riscv is broken.
3. Downloading the `openssl-libs-${{runner.os}}` artifact in `build_sqlite`, and passing correct `-I` and `-L` flags to make the compiler/linker use the static `libcrypto.a` build.
4. Your test action changes to test SQLCipher builds.

davidmartos96:
@simolus3 I'm not able to change the target branch without creating a new PR. Are you able to change it as admin?

.github/workflows/main.yml
```diff
@@ -432,3 +441,29 @@ jobs:
...
+
+      - name: Flutter sqlcipher tests on macOS
+        if: runner.os == 'macos'
+        working-directory: examples/flutter_integration_tests
+        run: |
+          flutter config --enable-swift-package-manager
+          flutter test integration_test -Dsqlite3.sqlcipher=true
+          flutter test integration_test -Dsqlite3.sqlcipher=true -d macos
+
+      - name: sqlcipher Android emulator tests
```
simolus3, line 457: Booting an emulator is quite slow. I wonder if the Android emulator tests should be a separate job now, where we might spawn a single Android emulator only and use a `script:` that runs multiple `flutter test` calls with different build options in sequence?

sqlite3/hook/build.dart
```diff
@@ -60,15 +60,70 @@ ${usedSqliteSymbols.map((symbol) => '    $symbol;').join('\n')}
         }
 
         final isSqlcipher = input.userDefines['is_sqlcipher'] as bool? ?? false;
```
simolus3, line 62: I'm to blame for this, but there should be a comment here explaining that these options are internal, undocumented and only meant to be set via `tool/build_sqlite.dart`

sqlite3/hook/build.dart
```diff
@@ -60,15 +60,70 @@ ${usedSqliteSymbols.map((symbol) => '    $symbol;').join('\n')}
...
+            if (!await openSslCompileDir.exists()) {
+              throw StateError(
+                'Expected OpenSSL compiled directory at ${openSslCompileDir.path}',
+              );
+            }
+
+            openSslStaticLib = File(
+              p.join(
+                openSslCompileDir.path,
+                _getOpenSslLibFolderName(targetOS, targetArchitecture),
```
simolus3, line 104: It seems like we can move this out of the hook and into `build_sqlite.dart` by just passing the OS-specific subdirectory?

sqlite3/lib/src/hook/description.dart
```diff
@@ -394,11 +404,44 @@ extension type const CompilerDefines(Map<String, String?> flags)
     return CompilerDefines(entries);
   }
 
-  static CompilerDefines defaults(bool windows) {
+  static CompilerDefines defaults({
+    required OS targetOS,
+    required bool isSqlcipher,
```
simolus3, line 409: Why did we move this from `tool/build_sqlite.dart` to here? The build script had those extra definitions [here](https://github.com/simolus3/sqlite3.dart/blob/f9fe9751aa13b9b9a9fd30bda659975dc0fee7b0/tool/build_sqlite.dart#L113-L122)?

simolus3 (commented):
Thanks for the help! This is definitely helpful, I can still look at Windows and non-X64 Linux support before merging and releasing the `sqlcipher` branch.

sqlite3/lib/src/hook/description.dart
davidmartos96: My reasoning is that if someone wants to customize the build, they could do so, even if they need to rely on some hidden config options. 
For instance, if we move it outside, we would lose the customization part because we couldn't pass the "link with Common Crypto" define on Apple platforms conditionally

sqlite3/hook/build.dart
davidmartos96: I guess we could add an "openssl_lib_dir" option. But because it was a derivative of the "openssl_build_dir" value, I wanted to keep it simple. And the build dir is also used for the include directory.
Your proposed design would be to make is as dummy as possible?
openssl_lib_dir and openssl_include_dir?

.github/workflows/main.yml
davidmartos96: Good idea. Should we create a bash script for the ci tests? Or would you like placing the "&&"s in the workflow yaml directly?

davidmartos96:
> Thanks for the help! This is definitely helpful, I can still look at Windows and non-X64 Linux support before merging and releasing the `sqlcipher` branch.

I would appreciate if you look into the Windows and Linux Cross Compiling. I'm not very experienced with the build tools for those

.github/workflows/main.yml
simolus3: Directly in the yaml should work

sqlite3/hook/build.dart
simolus3: I missed that the include dir is a different one, I agree that we can keep it as-is then.

simolus3:
I'll merge this as-is since it's a clear improvement. Thanks for your help, I'll look into the cross-compiling issues too.

PR #384, opened 2026-07-01
fix: use deterministic string for download cache dirname instead of Object.hash()

floatinghotpot:
Each time when build, the hook dart will ignore the local cached lib binary and always attempt to download the lib binary from github. It fails to detect the local cached lib binary with wrong folder name generated from Object.hash(). 

After troubleshooting, I find that the root cause is Object.hash().

Object.hash() uses a per-isolate random seed in Dart, meaning the hashCode — and therefore the cache directory name — changes on every hook invocation. Previously downloaded binaries were never found on subsequent builds, forcing a re-download from GitHub each time.

Replace hashCode-based dirname with deterministic string concatenation of releaseTag, type basename, os name, and architecture name. This makes the cache directory name stable across Dart processes.

Example: download-sqlite3-3.3.3_sqlite3_android_arm64

### DIFF
--- a/sqlite3/lib/src/hook/assets.dart
+++ b/sqlite3/lib/src/hook/assets.dart
@@ -106,7 +106,13 @@ final class PrebuiltSqliteLibrary {
   /// this library.
   ///
   /// There's a test asserting that this is unique for all valid values.
-  String get dirname => 'download-${hashCode.toRadixString(16)}';
+  ///
+  /// Uses a deterministic name so the download cache survives across separate
+  /// hook invocations (different Dart processes). Previously this used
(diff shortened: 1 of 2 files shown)


simolus3 (commented):
Yikes! Thanks for the contribution, the download directories definitely should be stable.

A concern against the full name is that Windows has an aggressive limitations on path names by default I believe, that's also the reason we introduced hashes in the first place. But something that should be more stable and still unique would be to use say the first 8 chars of the sha256 hash. I have done that on [this branch](https://github.com/simolus3/sqlite3.dart/tree/refs/heads/fix-asset-hashcodes), I'll release that after tests complete.

simolus3:
Released with version `sqlite3: 3.3.4`.

PR #394, opened 2026-08-02
Resolve relative paths in additional_includes and additional_lib_directories

dariyooo:
With `source: source`, `path` is resolved against the pubspec, but `additional_includes` and `additional_lib_directories` are passed through as-is. `native_toolchain_c` then resolves them against its build directory (`.dart_tool/hooks_runner/shared/sqlite3/build/<checksum>/`), so relative entries can never point at anything useful and only absolute paths work today.

This resolves both against the pubspec, matching `path`. Absolute entries pass through unchanged. Not breaking, since relative values are currently unusable.

The base is read from `input.json['user_defines']['workspace_pubspec']['base_path']` because `HookInputUserDefines.path()` only resolves a single string.

Test added to `test/hook/description_test.dart`.

### DIFF
--- a/sqlite3/lib/src/hook/compile/description.dart
+++ b/sqlite3/lib/src/hook/compile/description.dart
@@ -17,6 +17,7 @@ import '../utils.dart';
 sealed class SqliteBinary {
   static SqliteBinary forBuild(BuildInput input) {
     final userDefines = input.userDefines;
+    final userDefinesBase = _userDefinesBase(input);
 
     PrecompiledFromGithubAssets fromGitHub(LibraryType type) {
       final pattern =
@@ -24,6 +25,20 @@ sealed class SqliteBinary {
           PrecompiledFromGithubAssets.defaultUrlPattern;
 
       return PrecompiledFromGithubAssets(type, urlPattern: pattern);
+    }
+
+    List<String> resolvedPaths(String key) {
+      final List<String> entries =
+          (userDefines[key] as List?)?.cast() ?? const [];
+      if (userDefinesBase == null) return entries;
+
+      return [
+        for (final entry in entries)
+          if (p.isAbsolute(entry))
+            entry
+          else
+            userDefinesBase.resolve(entry).toFilePath(),
+      ];
     }
 
     switch (userDefines['source']) {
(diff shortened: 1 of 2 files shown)


sqlite3/lib/src/hook/compile/description.dart
```diff
@@ -78,6 +92,22 @@ sealed class SqliteBinary {
...
     }
   }
+
+  /// The `pubspec.yaml` the user-defines were read from.
+  ///
+  /// Relative paths are resolved against this file, matching what
+  /// [HookInputUserDefines.path] does for `path`. That method only resolves a
+  /// single string, so list-valued options need the base itself.
+  static Uri? _userDefinesBase(BuildInput input) {
+    final userDefines = input.json['user_defines'];
```
simolus3, line 102: It looks like there is no better way to do this at the moment, but us crawling through undocumented JSON structures doesn't seem great. I've opened https://github.com/dart-lang/native/issues/3515 to ask for a proper API.

simolus3 (commented):
Note that these options are intended to be used internally only.

If you want to add extensions, is it a requirement for you to link those statically? Dynamically loading extensions through a separate asset is [much simpler](https://github.com/simolus3/sqlite3.dart/tree/main/sqlite3/example/custom_extension), doesn't require changes to this package, and allows us to refine the build in the future (exposing new options puts constraints on that).

dariyooo:
> If you want to add extensions, is it a requirement for you to link those statically? 

Sadly, I do not really see another option. l also thought about just compiling a complete sqlite dynamic library that has all symbols in it but I also do not see a way to load that.

simolus3:
Out of interest, what extension are you trying to load (if it's public)? Most extensions can be configured to call SQLite APIs through function pointers, which means you don't need SQLite and the extension to be linked together statically.

dariyooo:
It is not really public but it is [this one](https://www.sqlite.org/zipvfs/doc/trunk/www/readme.wiki).

simolus3:
> If you already have an application that is built using the public-domain "sqlite3.c" source file, then the easiest way to get up and running with ZIPVFS it to substitute the "sqlite3-zipvfs.c" source file included with the ZIPVFS source tree in place of the public-domain "sqlite3.c" source file and recompile.

Would it be an option for you to vendor `sqlite3-zipvfs.c` with your app? This package can compile [custom SQLite builds](https://pub.dev/documentation/sqlite3/latest/topics/hook-topic.html#custom-sqlite-builds) if you provide the path to the sqlite library to compile.

This PR is effectively blocked on https://github.com/dart-lang/native/issues/3515.

dariyooo:
> Would it be an option for you to vendor sqlite3-zipvfs.c with your app? This package can compile [custom SQLite builds](https://pub.dev/documentation/sqlite3/latest/topics/hook-topic.html#custom-sqlite-builds) if you provide the path to the sqlite library to compile.

Thank you, this is what I am going to do for now.
But it would be nice to have this anyways as I am compiling some other extensions too. This is very cumbersome to do as I have to setup an Action and compile each of them individually and then download the binaries to load as extensions.

simolus3:
I have merged your changes after upgrading `package:hooks` to use a stable API in f02e4a29e7fc599cdec6698b058bcdf97b8f5237.

PR #397, opened 2026-08-03
Avoid callbacks when opening connection pools

### LINKED ISSUE #396: sqlite3_connection_pool: support opening pools without executable-memory FFI callbacks
### Summary

`sqlite3_connection_pool` currently requires the Dart VM to allocate executable memory whenever a pool is opened, even when native update hooks are disabled. This causes release/AOT Flutter applications to abort on GrapheneOS when the per-app **Dynamic code loading from memory** restriction is enabled.

This came up while investigating [ente/ente#3436](https://github.com/ente/ente/issues/3436).

### Root cause

[`RawSqliteConnectionPool.open`](https://github.com/simolus3/sqlite3.dart/blob/e09a2292ddcefc6bc89941ac76b8fe5098089521/sqlite3_connection_pool/lib/src/raw.dart#L181-L244) unconditionally creates an initializer with:

```dart
NativeCallable<Pointer<InitializedPool> Function()>.isolateLocal(...)
```

Dart implements `NativeCallable` by allocating an FFI callback trampoline page and changing it from writable to executable ([Dart VM source](https://github.com/dart-lang/sdk/blob/main/runtime/vm/ffi_callback_metadata.cc#L127-L225)). GrapheneOS denies that `mprotect` operation when in-memory dynamic code loading is restricted, and the Dart VM terminates the process. The failure is fatal and cannot be caught by the application.

Setting `PoolConnections(enableNativeUpdateHooks: false)` does not avoid this requirement: It disables the SQLite update hooks, but the initializer `NativeCallable` is still constructed before `pkg_sqlite3_connection_pool_open` is called.

### Requested behavior

Could pool initialization support an AOT path that does not dynamically allocate executable callback trampolines?

Two possible directions appear to be:

1. Use a statically compiled `Pointer.fromFunction` initializer backed by carefully scoped isolate-local initialization state.
2. Adjust the native API so the initialized connections/options can be passed without a Dart callback.

Ideally, `open` and `openAsync` would retain their current multi-isolate, exception propagation and cleanup behavior while working when new anonymous executable mappings are prohibited.

simolus3:
`NativeCallable` and `Pointer.fromFunction` are implemented by the Dart VM allocating an FFI stub in a page that is then remapped as executable. This is forbidden on some platforms like GrapheneOS.

To avoid that issue, this splits the `pkg_sqlite3_connection_pool_open` function into three parts:

1. `pkg_sqlite3_connection_pool_open` locks the global mutex of active pools and either returns a handle to the opened pool if it exists already, or indicates that the caller is responsible for initialization.
2. `pkg_sqlite3_connection_pool_initialize` takes the role of what used to be a callback, turning an initialized pool into a pool client and unlocking.
3. `pkg_sqlite3_connection_pool_close_uninitialized` drops an uninitialized handle without creating a pool, used on exceptions.

Closes https://github.com/simolus3/sqlite3.dart/issues/396.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -155,7 +155,7 @@ jobs:
       working-directory: sqlite3_connection_pool
       run: |
         dart build cli --target test/pool_test.dart -o out
-        systemd-run --user --pty --property=MemoryDenyWriteExecute=yes ./out/bundle/bin/pool_test
+        systemd-run --user --pipe --wait --collect --property=MemoryDenyWriteExecute=yes --property=RuntimeMaxSec=10 ./out/bundle/bin/pool_test
         rm -rf out
 
     - name: Enable sqlite3mc
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -149,6 +149,15 @@ jobs:
         dart run.dart --sanitizer=msan
       working-directory: native_tests/
 
+    # Regression test for https://github.com/simolus3/sqlite3.dart/issues/396
+    - name: Test sqlite3_connection_pool with MemoryDenyWriteExecute
+      if: runner.os == 'Linux'
+      working-directory: sqlite3_connection_pool
+      run: |
+        dart build cli --target test/pool_test.dart -o out
+        systemd-run --user --pty --property=MemoryDenyWriteExecute=yes ./out/bundle/bin/pool_test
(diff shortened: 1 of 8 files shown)


simolus3:
Merged in f1cd43538b3cf379fa4474628a8d18412cd8b252.

PR #403, opened 2026-08-16
Fix async OPFS offsets beyond 2 GiB

jpohhhh:
## Summary

The async OPFS VFS transports `WasmFile` offsets, truncate sizes and `xFileSize` results through three signed 32-bit fields in `MessageSerializer`. Once a database exceeds 2 GiB, those values wrap negative. Reads then target incorrect offsets and SQLite reports generic SQL errors/corruption despite valid data.

This changes the three protocol fields to signed 64-bit values encoded as low-uint32/high-int32 pairs. That representation works in dart2js (where `ByteData.setInt64` is unsupported) while preserving every JavaScript-safe integer. SQLite's maximum practical file size is below the 53-bit exact-integer limit.

The protocol version is bumped and filename metadata moves from byte 12 to byte 24.

## Reproduction

A production encrypted OPFS database failed deterministically just after crossing the boundary:

- `page_count = 263872`
- `page_size = 8192`
- file size = `2161639424` bytes
- ordinary INSERT returned `SQLITE_ERROR`
- reads and `PRAGMA quick_check` then reported unrelated page failures
- the identical workload on a smaller database passed

With this patch, the same production workload completes and reopens cleanly at:

- 2,112 transactions / 415,027 inserted rows
- `page_count = 281492`
- `page_size = 8192`
- file size = `2305982464` bytes
- `PRAGMA integrity_check` passes
- FTS5 integrity check passes
- close/reopen integrity checks pass

## Tests

Added browser serializer tests around:

- signed 32-bit maximum
- 2 GiB boundary
- 4 GiB boundary
- SQLite's maximum practical database size
- negative sentinel values
- filename metadata adjacent to widened flags

Validated under both dart2js and dart2wasm Chrome test platforms.

### DIFF
--- a/sqlite3/lib/src/wasm/vfs/async_opfs/sync_channel.dart
+++ b/sqlite3/lib/src/wasm/vfs/async_opfs/sync_channel.dart
@@ -3,7 +3,7 @@ import 'dart:typed_data';
 
 import '../../js_interop.dart';
 
-const protocolVersion = 1;
+const protocolVersion = 2;
 const asyncIdleWaitTimeMs = 150;
 const asyncIdleWaitTime = Duration(milliseconds: asyncIdleWaitTimeMs);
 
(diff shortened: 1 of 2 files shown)


jpohhhh:
I'm a bit annoyed, AI wrote up this PR and sent it without me asking for it to.

That being said, it is correct. 

TL;DR:
- I was stress testing my new SQLite-backed store on web
- importing, into a Flutter not-WASM web app, from Supabase.
- Database ends up at ~3.6 GB w/native Draft 
- I got odd failures, after about 8-10 hours of work & testing and futzing around if it was FTS or a large row, AI found the int32 thing
- can't use int64 b/c that's too big for Flutter not-WASM on web
- after this patch, import finishes cleanly both in an E2E test and in the web app itself

simolus3:
> can't use int64 b/c that's too big for Flutter not-WASM on web

This actually kind of works, we can bind to [setBigInt64](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/setBigInt64) directly and invoke `BigInt` as a JavaScript function to convert between Dart ints (as doubles) and big integers in JavaScript.

I've done that in the linked commit and copied your agent's testcase with minor fixes.

PR #406, opened 2026-09-05
Hook: allow source_<os> user-defines and path-like names for source: system

santoshakil:
Two small additions to the build hook for apps that ship SQLite inside one of their own native libraries.

With `source: system`, a `name` is currently always turned into a platform file name (`libfoo.dylib`, `libfoo.so`, ...). That does not work for a library that is bundled as a framework on iOS, where the loader needs `foo.framework/foo` (or `@rpath/...`). Names with a directory component are now passed to the loader unchanged, absolute paths keep working as before and plain names are still mapped like today.

`source` and `name` can now also be given per operating system as a map (`android: system`, `default: sqlite3` and so on; the existing `name_<os>` keys keep working). Our case, the Immich mobile app: a Rust library in the app already links SQLite with the compile options from this package's binaries and exports the `sqlite3_*` symbols, so on Android and iOS we point `package:sqlite3` at that library (one copy of SQLite in the app) while `flutter test` on the host keeps the default download.

Tests added in `test/hook/description_test.dart` for the path-like names (framework partial, `@rpath/...`, plain name still mapped) and for the per-OS maps (the iOS entry vs `default`, and a map without an entry for the target OS). The existing download url test now passes a code config since the source lookup needs the target OS. Docs in `doc/hook.md` and a changelog line.

Tried it in our app on an iPad simulator: the manifest entry becomes `["system", "immich_core_ffi.framework/immich_core_ffi"]`, no `sqlite3.framework` in the bundle, and `sqlite3_libversion` resolved through the sqlite3 asset lives in the same loaded image as our Rust library.

### DIFF
--- a/sqlite3/doc/hook.md
+++ b/sqlite3/doc/hook.md
@@ -141,21 +141,23 @@ hooks:
       # Names like `@rpath/libsqlite3.dylib` are supported as well.
 ```
 
-### Per-OS sources
+### Per-OS sources and names
 
-Like `name`, the `source` key can be overridden for specific operating systems with `source_$os`
-keys. This allows using a library from the system (or from the app) on some platforms while
-keeping the default binaries on others:
+`source` and `name` can also be maps from target operating systems to values. The `default`
+entry applies to operating systems not listed in the map (without it, the key is treated as
+unset for those):
 
 ```yaml
 hooks:
   user_defines:
     sqlite3:
-      source: sqlite3 # default, used for platforms not listed below
-      source_android: system
-      source_ios: system
-      name_android: my_native_lib
-      name_ios: my_native_lib.framework/my_native_lib
+      source:
+        android: system
+        ios: system
+        default: sqlite3
+      name:
+        android: my_native_lib
+        ios: my_native_lib.framework/my_native_lib
 ```
 
 ## Custom SQLite builds
--- a/sqlite3/doc/hook.md
+++ b/sqlite3/doc/hook.md
@@ -113,9 +113,10 @@ When using `source: system`, you can also add a `name` key to customize the name
 library.
 For instance, `name: sqlcipher` would load `libsqlcipher.dylib` on macOS, `libsqlcipher.so` on Linux
 and `sqlcipher.dll` on Windows.
-To further customize the name for specific operating systems, you can use `name_$os` keys, where `$os`
-is the [target operating system](https://pub.dev/documentation/code_assets/latest/code_assets/OS-class.html)
-passed to hooks:
+To further customize the name for specific operating systems, `name` can be a map with an entry per
+[target operating system](https://pub.dev/documentation/code_assets/latest/code_assets/OS-class.html)
+passed to hooks. The `default` entry applies to operating systems not listed in the map, and the
+package's usual default (`sqlite3`) is used for them if there is no `default` entry either:
 
 ```yaml
 hooks:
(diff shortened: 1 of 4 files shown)


sqlite3/doc/hook.md
```diff
@@ -127,6 +127,37 @@ hooks:
...
+
+Like `name`, the `source` key can be overridden for specific operating systems with `source_$os`
+keys. This allows using a library from the system (or from the app) on some platforms while
+keeping the default binaries on others:
+
+```yaml
+hooks:
+  user_defines:
+    sqlite3:
+      source: sqlite3 # default, used for platforms not listed below
```
simolus3: I feel like a syntax based on subkeys might be cleaner, can we do something like this:

```yaml
source:
  android: system
  ios: system
  default: sqlite3
name:
  android: my_native_lib
  ios: my_native_lib.framework/my_native_lib
```

simolus3 (commented):
I really wish hooks had a way to configure this out of the box, but since they don't I feel it makes sense to adopt something in the package itself 👍

sqlite3/doc/hook.md
santoshakil: done, both keys take the map now with a default entry, the old name_<os> keys still work.

sqlite3/doc/hook.md
```diff
@@ -127,6 +127,39 @@ hooks:
       name: sqlite3
```
simolus3: Could you also update this to the new syntax?

```suggestion
     name:
       windows: winsqlite3
       default: sqlite3
```

sqlite3/lib/src/hook/compile/description.dart
```diff
@@ -53,10 +76,10 @@ sealed class SqliteBinary {
       case 'test-sqlcipher':
         return PrecompiledForTesting(LibraryType.sqlcipher);
       case 'system':
-        final osSpecificNameKey = 'name_${input.config.code.targetOS.name}';
+        final osSpecificNameKey = 'name_${targetOS.name}';
```
simolus3, line 80: Please add a comment here mentioning that this is for backwards compatibility before os-specific subkeys were supported.

sqlite3/doc/hook.md
```diff
@@ -127,6 +127,39 @@ hooks:
...
+another native library of the app already links SQLite and exports its symbols:
+
+```yaml
+hooks:
+  user_defines:
+    sqlite3:
+      source: system
+      name_ios: my_native_lib.framework/my_native_lib
+      # Names like `@rpath/libsqlite3.dylib` are supported as well.
+```
```
simolus3, line 148: Instead of adding a new section here, update `With custom library names` above to reflect the new syntax.

sqlite3/doc/hook.md
```diff
@@ -127,6 +127,39 @@ hooks:
...
+
+### Per-OS sources and names
```
simolus3: After merging the previous section into the existing one, this can probably be a shorter note explaining that this also works for `name`.

sqlite3/doc/hook.md
```diff
@@ -127,6 +127,39 @@ hooks:
...
+
+`source` and `name` can also be maps from target operating systems to values. The `default`
+entry applies to operating systems not listed in the map (without it, the key is treated as
+unset for those):
```
simolus3: > without it, the key is treated as unset for those

While technically correct, I think it might be more helpful to say instead that the default from the package would be applied in that case.

simolus3 (approved):

PR #409, opened 2026-09-12
Await in-flight batch in IndexedDbFileSystem.flush()

### LINKED ISSUE #408: IndexedDbFileSystem.flush() does not await a batch already in flight (regression in 3.4.0)
## Summary

`IndexedDbFileSystem.flush()` returns without awaiting when a write batch is already in flight. Its own doc comment promises the opposite:

> Waits for all pending operations to finish, then completes the future.
> Each call to [flush] will await pending operations made _before_ the call.

This is a regression in 3.4.0. In 3.3.4 `flush()` queued a marker behind the running batch and awaited it; the marker was removed in `11be8acb` ("Optimize indexeddb flush on idle").

## The change

```diff
-  Future<void> flush() async {
-    final item = _submitWorkFunction((_) async {}, 'flush');
-    _startWorkingIfNeeded(isImplicit: false);
-    await item;
+  Future<void> flush() {
+    return _startWorkingIfNeeded(isImplicit: false);
   }
```

The same commit made `_startWorkingIfNeeded` `async` and `await` the writes, which makes the *idle* path a genuine fence and removes the marker's overhead. But the busy path is now a no-op:

```dart
Future<void> _startWorkingIfNeeded({required bool isImplicit}) async {
  if (isImplicit && !_writeAutomatically) return;
  if (!_isWorking && _pendingWork.isNotEmpty) {   // ← skipped while a batch runs
    ...
    await _asynchronous._performWrites(items).whenComplete(...);
  }
}
```

With `writeAutomatically: true` (the default), `xWrite` → `_submitWork` → `_startWorkingIfNeeded(isImplicit: true)` sets `_isWorking` synchronously, so right after any write the flag is already set. A `flush()` at that moment takes no branch and its future completes immediately, before the IndexedDB transaction commits.

## Reproduction

Any write followed by a flush:

```dart
final fs = await IndexedDbFileSystem.open(dbName: 'repro');
// ... open a database on this VFS and INSERT something ...
await fs.flush();   // returns before the IndexedDB transaction commits on >= 3.4.0
```

Measured in headless Chrome: on 3.5.2 `flush()` completed after **0** event-loop turns; on 3.3.3 it waited **7**.

Version boundary, checked against the source at each tag:

| version | `flush()` body | fence? |
|---|---|---|
| 3.3.3, 3.3.4 | `_submitWorkFunction(() {}, 'flush')` | yes |
| 3.4.0 … 3.5.2 | `_startWorkingIfNeeded(isImplicit: false)` | only when idle |

## Why it matters

`flush()` is the only way to force a transaction. `open()`'s doc says writes are "asynchronously written to IndexedDB without any durability guarantees. You can invoke [flush] to force a transaction", and the 3.4.0 changelog says `writeAutomatically: false` users "need to call `flush` manually" — for those users a flush that does not fence means the write may not have happened at all yet.

For default users the window is small (the VFS streams writes continuously, so what is missed is the batch in flight), but it is exactly the window an app closes in when it flushes on `visibilitychange` before the tab goes away.

## Suggested fix

Keep the idle optimisation, restore the marker only when busy:

```dart
Future<void> flush() {
  if (!_isWorking) return _startWorkingIfNeeded(isImplicit: false);
  // A batch is already running: queue a marker behind it and await that.
  return _submitWorkFunction((_) async {}, 'flush');
}
```

`_FunctionWorkItem` does not override `insertInto`, so it always lands at the tail of `_pendingWork` and is never merged or skipped. When the running batch finishes, `whenComplete` restarts work and the next batch carries every write submitted before the flush, plus the marker last — so the marker's completer resolves only after they are committed. This is what `close()` already does.

Happy to open a PR for this.

Found while adding an explicit persistence call to a package that uses this VFS for a vector store (DenisovAV/flutter_gemma#507).

DenisovAV:
Fixes #408.

`11be8acb` made `flush()` return `_startWorkingIfNeeded(isImplicit: false)` directly. That is a genuine fence when the worker is idle — `_startWorkingIfNeeded` now awaits `_performWrites` — and it removed the marker's overhead. But while a batch is already being written, `_startWorkingIfNeeded` takes no branch and its future completes immediately, so `flush()` returns before that batch reaches IndexedDB. With `writeAutomatically: true`, `_isWorking` is set synchronously by the implicit start in `_submitWork`, so this is the ordinary state right after any write.

This keeps the idle fast path and restores the marker only when busy:

```dart
Future<void> flush() {
  if (!_isWorking) {
    return _startWorkingIfNeeded(isImplicit: false);
  }
  return _submitWorkFunction((_) async {}, 'flush');
}
```

Why the marker is sound: `_FunctionWorkItem` does not override `insertInto`, so it always lands at the tail of `_pendingWork` and is never merged or skipped. When the running batch finishes, `whenComplete` restarts work, and the next batch carries every write submitted before the flush plus the marker last. With `writeAutomatically: false` the running batch can only have been started explicitly, so its restart also uses `isImplicit: false` and proceeds. This is the same shape `close()` already uses.

### On testing

I haven't added a test, and want to be upfront about why rather than add one that passes for the wrong reason. The obvious check — flush, then read the data back through a fresh `IndexedDbFileSystem` or `AsynchronousIndexedDbFileSystem` — doesn't discriminate: an IndexedDB transaction that overlaps a pending readwrite transaction on the same store waits for it, so the read sees the data whether or not `flush()` actually awaited. A test that asserts on the number of event-loop turns before `flush()` resolves would catch it (it reads 0 on 3.5.2 and 7 on 3.3.3 in headless Chrome), but that is fragile. If you have a preferred way to observe work-queue state from a test, I'm happy to add one.

Verified: `dart analyze` clean on the changed file, `dart format` unchanged.

### DIFF
--- a/sqlite3/CHANGELOG.md
+++ b/sqlite3/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 3.6.0 (unreleased)
 
+- Fix `IndexedDbFileSystem.flush()` completing before a write batch already in progress was persisted.
 - Hooks: Support relative paths in `additional_includes` and `additional_lib_directories`.
 - Hooks: Support multiple source files to compile with `source: source`.
 - Hooks: `source` and `name` accept a map with per-OS entries and a `default` (e.g. `source: {android: system, default: sqlite3}`). Names containing a path (like `foo.framework/foo`) are passed to the loader as-is with `source: system`.
--- a/sqlite3/lib/src/wasm/vfs/indexed_db.dart
+++ b/sqlite3/lib/src/wasm/vfs/indexed_db.dart
@@ -607,7 +607,16 @@ final class IndexedDbFileSystem extends BaseVirtualFileSystem {
   /// Operations started after this [flush] call will not be awaited by the
   /// returned future.
   Future<void> flush() {
-    return _startWorkingIfNeeded(isImplicit: false);
+    if (!_isWorking) {
+      return _startWorkingIfNeeded(isImplicit: false);
+    }
+
+    // A batch is already being written, so `_startWorkingIfNeeded` would take
+    // no branch and complete immediately — before that batch, or anything
+    // queued behind it, reaches IndexedDB. Queue a no-op item instead: it lands
+    // at the tail of `_pendingWork` and completes only once the work ahead of
+    // it has been written, which is the guarantee documented above.
+    return _submitWorkFunction((_) async {}, 'flush');
   }
 
   @override


simolus3 (commented):
Thanks for the PR. I agree the current state is buggy, but we can likely clean up the fix a bit.

Instead of adding a separate branch here, can we make `flush()` return the future associated with a prior work item? The easiest way to do that would be to replace `bool _isWorking` with a `Future<void>? _currentWork`.

In `flush()`, we then await `_currentWork` or `_startWorkingIfNeeded` if it's unset. This avoids scheduling a bogus work item, which also starts an IndexedDB transaction (so it's not free).

